PR #15, opened 2018-10-30
Improve example documentation

theZiz:


examples/asynccopy/asynccopy.cpp
```diff
@@ -1,18 +1,42 @@
...
+/** \file asynccopy.cpp
+ *  \brief Asynchronous bluring example for LLAMA using ALPAKA.
+ */
+
 #include <iostream>
 #include <utility>
 #include <list>
 
+/// defines whether cuda shall be used or openmp
 #define ASYNCCOPY_CUDA 0
```
ax3l: I usually prefer to put such controls in a
```C++
#ifndef VAR
#define VAR default
#endif 
```

block so one can control them without warnings externally.

theZiz: You are right, I added those so I can quick change them in file while debugging / benchmarking, but I should go your way. Would you like to open an issue for this? Then I will tackle this soon. :wink:

ax3l: My pleasure, done in #16

PR #19, opened 2019-03-05
Do not magically expect an `alpaka` somewhere in the source tree

bertwesarg:


psychocoderHPC:
I realized that llma has no dependencies to alpaka. Therefore the path in this example is pointing somewhere into the nirvana. 
We should remove the hardcoded path completly. Than the user has several whys to point to alpaka.

- extent the environment variable `CMAKE_PREFIX_PATH`
- set the environment variable `ALPAKA_ROOT`

bertwesarg:
> I realized that llma has no dependencies to alpaka.

Correct, its only used in the examples.

> Therefore the path in this example is pointing somewhere into the nirvana.

Mostly correct, I consider this initial cache value an educated guess and is vary handy. As long as the user is able to overwrite this value by setting some of the below variables, I'm in favor to keep this initial cache value.

> We should remove the hardcoded path completly. Than the user has several whys to point to alpaka.
> 
>     * extent the environment variable `CMAKE_PREFIX_PATH`
> 
>     * set the environment variable `ALPAKA_ROOT`

psychocoderHPC:
But as you saw on your own environment this can lead very soon into undefined behavior. We should stay with standard ways to move locations into CMake if the dependency is not provided via submodule or subtree.

bertwesarg:
> But as you saw on your own environment this can lead very soon into undefined behavior. We should stay with standard ways to move locations into CMake if the dependency is not provided via submodule or subtree.

Point taken.

If we add `alpaka` as an submodule, than I think it should also live under `examples`. But without knowing, which version of `alpaka` should be used, I think we wont circumvent adding it as an submodule at all.

psychocoderHPC:
Mhhh I am not sure if we should add the alpaka monster as submodule but maybe it is a way to go.
But yes if we add this than only into the examples folder.

theZiz:
Please, no submodules... :scream: 

I would suggest to leave it this way or as @psychocoderHPC said to remove the path completely and to have the user point to an alpaka installation explicitly.

ax3l:
I agree that submodules are not ideal due to their weird behavior for non-git-superheros on switch of branches, etc.

I would also remove the educational guess of an Alpaka path, since there is a very well defined way how CMake finds packages. E.g., in CMake 3.12+ one only has to `export alpaka_ROOT=<Path>` and all is set :)

https://cmake.org/cmake/help/v3.12/command/find_package.html

P.S.: Since LLama is still CMake 3.3+, we should set the [`CMP0074` policy](https://cmake.org/cmake/help/v3.12/policy/CMP0075.html):

```cmake
if(POLICY CMP0074)
    cmake_policy(SET CMP0074 NEW)
endif()
```

https://github.com/ComputationalRadiationPhysics/alpaka/pull/751

bertwesarg:
> E.g., using CMake 3.12+ one only has to `export alpaka_ROOT=<Path>` and all is set :)
> 
> https://cmake.org/cmake/help/v3.12/command/find_package.html
> 

Is this really case sensitive? My understanding was, that `<PackageName>` is case insensitive! But how to name `<PackageName>_ROOT` than?

ax3l:
Afaik, package names in CMake are always case sensitive and come from the definition of your `project(name)` that then translates in a `nameConfig.cmake` when consistently installed.

bertwesarg:
Do you know out of your head, how project names with `-` are translated to the env name? There is no statement in https://cmake.org/cmake/help/v3.13/envvar/PackageName_ROOT.html, and https://cmake.org/cmake/help/v3.13/command/project.html also does not mention any restrictions to project names!

ax3l:
Depends also a bit how you name your `*Config.cmake` file, but likely keeps the `-` in the name.

Nevertheless, CamelCase is the community standard for all kinds of targets and projects: https://cliutils.gitlab.io/modern-cmake/chapters/install/installing.html

bertwesarg:
Whatever, back to the topic: No submodules, thus we prefill `ALPAKA_ROOT` just with `""` and then later we switch to the CMP0074 policy.

ax3l:
Yes, feel free to update that way.

psychocoderHPC:
Yes please let the user provide the location.

ax3l:
Yes, that's why he wants to update to "" by default, which will not be found. (We can also put "ALPAKA-NOTFOUND" which is interpreted as `false` by CMake.

bertwesarg:
Something like this?

examples/CMakeLists.txt
```diff
@@ -34,8 +34,10 @@ endif()
 # Alpaka examples #
 ###################
 
-set(ALPAKA_ROOT "${CMAKE_CURRENT_LIST_DIR}/../../../alpaka/" CACHE STRING  "The location of the alpaka library")
-LIST(APPEND CMAKE_PREFIX_PATH "${ALPAKA_ROOT}")
+set(ALPAKA_ROOT "ALPAKA-NOTFOUND" CACHE STRING  "The location of the alpaka library")
+if(NOT ALPAKA_ROOT STREQUAL "ALPAKA-NOTFOUND")
```
psychocoderHPC: I would say this must be `if(ALPAKA_ROOT)` only. I  will check it.

bertwesarg: Thanks, it took me 5 minutes, to figure out how to compare strings in CMake!

psychocoderHPC:
I would say we should remove the CMake variable `ALPAKA_ROOT` fully. The reason for that is that the common way is to use `alpaka_DIR` or extend the environment variable `CMAKE_PREFIX_PATH` or with the new policy CMP0074 to define the environment variable `alpaka_ROOT`

The self-defined variable will not fit into those concepts.

examples/CMakeLists.txt
ax3l: but... I linked it [above](https://github.com/ComputationalRadiationPhysics/llama/pull/19#issuecomment-474755430) :)

PR #20, opened 2019-03-05
Support `operator[]` to access virtual datum

bertwesarg:


theZiz:
I don't care it is a further, **optional** way of accessing a virtual datum. I like the `operator()` because of the possibility to give more than one parameter without the need to pass an intermediate type, which packs a multidimensional coordinate.

bertwesarg:
> lgtm, thx

thanks, time to merge?

PR #22, opened 2019-03-06
Carry the DatumDomain type in the StubType as type member `type`

bertwesarg:
This allows to pass the StubType to the external container, but also work
later on that type again with LLAMA by only having a container object.

### DIFF
--- a/documentation/pages/user/api.rst
+++ b/documentation/pages/user/api.rst
@@ -134,6 +134,7 @@ Datum domain
 
 .. doxygentypedef:: llama::StubType
    :project: LLAMA
+   :members:
 
 .. doxygentypedef:: llama::GetDatumElementType
    :project: LLAMA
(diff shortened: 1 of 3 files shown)


bertwesarg:
One question:

It looks like the most types in LLAMA are camel case, like `Mapping`, `DatumDomain`, and `UserDomain`. I.e., see `VirtualView`. `AoS`, ...

Should I have named the member type `DomainType`?

ax3l:
`type` as public typedef is quite common in the STL and boost, so that looks good for an inner type.

bertwesarg:
> `type` as public typedef is quite common in the STL and boost, so that looks good for an inner type.

My point is, that other inner types of LLAMA are in CamelCase. Thus the question is, should this one be considered part of the LLAMA convention or STL?

PR #23, opened 2019-03-19
N-Body: Need to use the block-level thread index to copy into the shared memory

bertwesarg:


bertwesarg:
@theZiz please check, @MarkusVelten please check, if this is the same we came up with

MarkusVelten:
@bertwesarg looks fine to me, it is just like what we worked out before

examples/nbody/nbody.cpp
```diff
@@ -211,6 +211,10 @@ struct UpdateKernel
             threads
         >;
 
+        auto threadBlockIndex  = alpaka::idx::getIdx<
```
psychocoderHPC: there is a space too much before `=`

bertwesarg: Seriously ;-)

psychocoderHPC: Yes the space was jumping directly in my face and it takes me 5min to get the additional space out of my head^^

bertwesarg: Than better do not look into this file at all before you apply my da3672f.

psychocoderHPC: ohh most of this was fine and following our [coding guide lines](https://github.com/ComputationalRadiationPhysics/contributing/blob/master/codingGuideLines/cpp.md) but I see we have not mentioned it somewhere.

Should I update your PR that you must not waste time into this. The problem is that no linter is supporting templates and we can therefore not do it automatically :-(

bertwesarg: I more or less choose /one/ arbitrary style used in this file and applied it.

Except for `&`, `&&`, and `*` I think I pretty much nailed it, didn't I?

bertwesarg: I pushed a second try, fell free to make your fixes on top of the 9b2473e7474ac4ab5b229e42c60b070611ac0ee0, so that I can better see, where I still was wrong. Thanks.

psychocoderHPC: Thanks for your update. I fixed only 3 or 4 small spaces issues. Did you use a tool for the reformatting or did you do it by hand? If you use a tool: which?

psychocoderHPC: ohh wait could you please mark on the right side the checkbox `Allow edits from maintainers.` else I can not push to your PR.

bertwesarg: no tool! but please rebase if you push!

examples/nbody/nbody.cpp
```diff
@@ -303,56 +303,56 @@ struct MoveKernel
...
         );
 
         LLAMA_INDEPENDENT_DATA
         for ( auto pos = start; pos < end; ++pos )
-            particles( pos )( dd::Pos() ) += particles( pos )( dd::Vel() ) * ts;
+            particles( pos )( dd::Pos( ) ) += particles( pos )( dd::Vel( ) ) * ts;
     }
 };
 
-int main(int argc,char * * argv)
```
bertwesarg: @psychocoderHPC  either you missed this, or the one in the real `main` was an error from your side (its `**` there). So what is the correct one?

ax3l: Of course, only
```C++
int main(
    int argc,
    char * argv[]
)
{
    // ...
}
```
is the right signature ;-D

bertwesarg: I'm fine with this too. But what about the general case of _multi pointer_ types? Are there spaces around each of them or are they packed?

psychocoderHPC: Ohh I missed that. `**` should be packed but I am also fine with `* x[]`

bertwesarg: What about `const` than: `int * const * foo;`?

psychocoderHPC: I never saw a const in the signature of the main function. I would not add `const`.
But if const is needed `int * const * foo` would fit

bertwesarg: My question was not limited to `main`!

ax3l: :ok_hand:

PR #21, opened 2019-03-05
Allow to use datum tag objects to access members

bertwesarg:

PR #25, opened 2019-03-22
datum tag objects to access members

psychocoderHPC:
This PR implements the same behavior than #21 but without removing the usage of universal references.

Allow using an instance of a datum domain to access members.

Co-authored-by: Bert Wesarg <bert.wesarg@tu-dresden.de>

@bertwesarg could you please have a look to this PR.


bertwesarg:
Quoting @theZiz:
> lgtm, thx

for me too ;-)

PR #26, opened 2019-03-28
Array: multiple fixes in push_{front,back}

bertwesarg:
@theZiz @psychocoderHPC please review

### DIFF
--- a/include/llama/Array.hpp
+++ b/include/llama/Array.hpp
@@ -100,7 +100,7 @@ struct Array
     pop_front() const
     -> Array<
         T,
-        count-1
+        count - 1
     >
     {
         Array<
@@ -120,7 +120,7 @@ struct Array
     pop_back() const
     -> Array<
         T,
-        count-1
+        count - 1
     >
     {
         Array<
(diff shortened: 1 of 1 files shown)


include/llama/Array.hpp
```diff
@@ -142,15 +142,15 @@ struct Array
...
     >
     {
         Array<
             T,
-            count+1
+            count + 1
         > result;
         for( std::size_t i = 0; i < count - 1; i++ )
-            result.element[ i + 1 ] = element[ i ];
+            result.element[ count - i ] = element[ count - 1 - i ];
```
psychocoderHPC: Why do you copy from back to front? IMO only the loop was wrong and must be `for( std::size_t i = 0; i < count; i++ )` since the array has `count + 1` elements.

psychocoderHPC: I would say even in you variant you miss one element.

bertwesarg: Because I considered it an in-place copy. But its not, thus we can keep the forward copy. Will fix.

PR #31, opened 2019-11-11
UpdateKernel: add synchronization

MarkusVelten:
Synchronization to ensure that correct data will be shared independent of blockSize (e.g. when blockSize > Warp size (32) on CUDA devices since only threads within Warp run simultaneously)

### DIFF
--- a/examples/nbody/nbody.cpp
+++ b/examples/nbody/nbody.cpp
@@ -275,7 +275,10 @@ struct UpdateKernel
                 pos2 + threadIndex < end2;
                 pos2 += threads
             )
+            {
                 temp( pos2 + threadBlockIndex ) = particles( start2 + pos2 + threadBlockIndex );
+            }
+            syncBlockThreads( acc );
 #endif // NBODY_USE_SHARED
             LLAMA_INDEPENDENT_DATA
             for ( auto pos2 = decltype( end2 )( 0 ); pos2 < end2; ++pos2 )
@@ -290,7 +293,10 @@ struct UpdateKernel
 #endif // NBODY_USE_SHARED
                         ts
                     );
-        };
+#if NBODY_USE_SHARED == 1
+            syncBlockThreads( acc );
+#endif // NBODY_USE_SHARED
+        }
     }
 };
 


examples/nbody/nbody.cpp
```diff
@@ -276,6 +276,9 @@ struct UpdateKernel
                 pos2 += threads
             )
                 temp( pos2 + threadBlockIndex ) = particles( start2 + pos2 + threadBlockIndex );
+
+            syncBlockThreads(acc);
```
bertwesarg:
```suggestion
            syncBlockThreads( acc );
```

examples/nbody/nbody.cpp
```diff
@@ -290,6 +293,8 @@ struct UpdateKernel
 #endif // NBODY_USE_SHARED
                         ts
                     );
+
+            syncBlockThreads(acc);
```
bertwesarg: I think we could avoid this barrier, in case we don't use the shared memory.

examples/nbody/nbody.cpp
```diff
@@ -276,6 +276,9 @@ struct UpdateKernel
                 pos2 += threads
             )
                 temp( pos2 + threadBlockIndex ) = particles( start2 + pos2 + threadBlockIndex );
```
ax3l: while we are on it: this line is ill-aligned :)
(four less spaces pls)

bertwesarg: why? its the statement of the for loop

ax3l: oh right, hard to see!

bertwesarg: thats true, maybe we should use `{}` here:

```diff
-                pos2 += threads
-            )
+                pos2 += threads )
+            {
                 temp( pos2 + threadBlockIndex ) = particles( start2 + pos2 + threadBlockIndex );
+            }
```

examples/nbody/nbody.cpp
theZiz: Is this second barrier at the end needed at all? I understand the first barrier: It may be that some warp starts working on shared memory which is not filled by another, but after the shared memory is filled, everything should be save?

examples/nbody/nbody.cpp
theZiz: Tbh I was wondering why this version worked at all without any `alpaka`namespaces (boy, I didn't use this kind of C++ for a way too long time, feel free to explain it too me :smile: ), but anyway: Imho the namespace of `syncBlockThreads` should be explicit: `alpaka::block::sync::syncBlockThreads( acc );`

examples/nbody/nbody.cpp
bertwesarg: Yes I think so, because some threads maybe already overwrite the shared memopry in the next round. Though we could move this barrier at the begin of the loop before line 272.

theZiz: Ah, I overlooked the outer loop because it ended with `};` instead of `}`. >.< You are right.

bertwesarg: One more reason to use `{}` for the loop body.

theZiz: What do you mean? In general a loop **should** end with `}`, not `};`, so my code has indeed a style fault, but if you have the comment of @ax3l in mind: I don't like one-line-loops to be encapsulated. :roll_eyes:

examples/nbody/nbody.cpp
```diff
@@ -290,6 +293,8 @@ struct UpdateKernel
 #endif // NBODY_USE_SHARED
                         ts
                     );
+
+            syncBlockThreads(acc);
         };
```
bertwesarg: while you are at it, please drop the ';' here

examples/nbody/nbody.cpp
ax3l: yeah, for all reasonable cases. nobody could anticipate what you wrote *here* :D

examples/nbody/nbody.cpp
ax3l:
```suggestion
        }
```

bertwesarg:
@psychocoderHPC please review, thanks

psychocoderHPC:
thx for the fix!

PR #35, opened 2020-06-24
Replaced png++ by stb_image

bernhardmgruber:
* Replaced png++ by stb_image
* some fixes for changes in alpaka API
* compilation fixes for nvcc 10.2 and 11.


examples/asynccopy/asynccopy.cpp
```diff
@@ -449,15 +452,10 @@ int main(int argc,char * * argv)
...
-                host( y, x )( px::R() ) = Element((*image)
-                    [ Y - kernelSize ]
-                    [ X - kernelSize ].red) / 255.;
-                host( y, x )( px::G() ) = Element((*image)
-                    [ Y - kernelSize ]
-                    [ X - kernelSize ].blue) / 255.;
-                host( y, x )( px::B() ) = Element((*image)
-                    [ Y - kernelSize ]
-                    [ X - kernelSize ].green) / 255.;
+                const auto* pixel = &image[(Y - kernelSize) * img_x + X - kernelSize * 3];
```
theZiz, line 455: I think the pixel calculation is wrong.
* `(Y - kernelSize) * img_x` should be `(Y - kernelSize) * img_x * 3` as each scanline is three times as big as `img_x`.
* `X - kernelSize * 3` should  be `(X - kernelSize) * 3` for the same reason.
At least if stb_image reads images as `rgbrgbrgbrgb`. But for `rrrrggggbbbb` the code would be wrong, too. :wink:

examples/asynccopy/asynccopy.cpp
```diff
@@ -612,24 +610,23 @@ int main(int argc,char * * argv)
     for (std::size_t y = 0; y < img_y; ++y)
         for (std::size_t x = 0; x < img_x; ++x)
         {
-            (*image)[ y ][ x ].red = host(
+            auto* pixel = &image[(y * img_x + x) * 3];
```
theZiz, line 613: Maybe just use this code for reading the pixel above?

examples/asynccopy/asynccopy.cpp
bernhardmgruber: Well observed! I have have missed that when I pulled the changes for this branch together!

bernhardmgruber: Should be fixed

theZiz:
Hm, I am not able to find alpaka. My old style CMakeLists.txt did work for me, but with your changes I have no idea which path I have to pass in which way to cmake. I checked out alpaka version 0.4.1, but this version doesn't seem to have `make install` implemented yet so I tried to pass the source directory directly but failed. :(

bernhardmgruber:
I checked out alpaka develop and installed it. LLAMA was able to pick that up, both on Windows and on Ubunut in WSL.

theZiz:
Ah, okay, I assumed some breaking changes in alpaka 0.5 as llama is explicitly searching for 0.4, I'll give it a shot

theZiz:
Okay, installing alpaka to `/usr/local` works. I guess I have to see myself how to pass custom module search paths to cmake.

However I wonder why you are shipping a version of stb_image with llama? Why not using the systems version?

bernhardmgruber:
I shipped the stb_image headers because I failed to find the package name on Ubuntu and figured, it's probably not available. On Windows I could run `vcpkg install stb`. Furthermore, AFAIK, stb_image does not provide CMake targets, so it's also not nice to package for package maintainers. Vcpkg with Visual Studio after installing stb just makes it globally available, but this solution does not work for other package managers. So the easiest solution for me seemd to ship the header.

theZiz:
Well, I can only speak for Debian:
https://packages.debian.org/buster/libstb-dev

psychocoderHPC:
@bernhardmgruber note: we use normally [PNGWriter](https://github.com/pngwriter/pngwriter) to build png files but it does not mean you need to use it in llama.

theZiz:
@psychocoderHPC The example also loads pngs. Does PNGWriter support this?

bernhardmgruber:
So I checked the Debian package and it only contains the headers, some libraries, but no CMake files. So `find_package(stb)` will still not work. I found out that vcpkg ships a `FindStb.cmake`, but then the vcpkg toolchain file needs to be specified at the cmake command line, which is annoying as well. And this only works with vcpkg. So either we ship the dependency (2 headers) or we omit the headers and any mentioning in the CMakeLists.txt and have compilation fail when the library is not installed.

theZiz:
Well, I guess it is okay for an example. :wink:

psychocoderHPC:
> @psychocoderHPC The example also loads pngs. Does PNGWriter support this?

Yes this is supported.

bernhardmgruber:
So we can merge?

psychocoderHPC:
Could it be that this PR break something? The CI has not passed.

bernhardmgruber:
The CI (that I added in the previous PR) did never pass so far, because LLAMA causes an internal compile error with MSVC and many tests depend on typeid and demangled symbols, which I created on Windows using clang-cl. I want to fix the tests at some point, but this is part of a later PR.

psychocoderHPC:
I closed my eyes and merged it even if the CI failed ^^

PR #36, opened 2020-06-24
Switching to C++17

bernhardmgruber:
So I had a call a while back with Michael on how I am allowed to proceed on LLAMA. I specifically asked for use of C++17 and he gave me greenlight. So here is the switch to C++17 in CMake.

I am creating this PR explicitely, because it may make sense to create a tag or release before the switch, just in case anyone wants to use LLAMA without C++17 support.


psychocoderHPC:
I will merge it, if we like to do a release with a commit before this PR, we can easily open a release branch.

PR #37, opened 2020-06-26
Fix tests on Windows/Ubuntu and make Ubuntu CI pass

bernhardmgruber:
* added a lot of platform specific handling for type_info names
* added some zero initialization of view storage to tests that calculate values
* fixed all tests to run successfully with clang-cl/Windows and g++/Ubuntu


bernhardmgruber:
I think this is as good as it gets. The Windows CI will not finish for now, as MSVC fails to compile LLAMA.

PR #38, opened 2020-06-26
More tests for tree mappings

bernhardmgruber:
- [x] requires merging #36 

* added deduction guide for `llama::tuple`
* more tests for tree mappings

### DIFF
--- a/include/llama/Tuple.hpp
+++ b/include/llama/Tuple.hpp
@@ -86,6 +86,8 @@ struct Tuple< T_FirstElement >
     FirstElement first;
 };
 
+template <typename... Elements>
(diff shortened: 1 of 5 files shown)


psychocoderHPC:
I merged #36

psychocoderHPC:
I will always look only to the ubuntu CI until windows is fixed. This PR broke the ubuntu CI

bernhardmgruber:
It seems like a compiler issue. g++7 fails, but g++9 compiles fine.

bernhardmgruber:
I was wrong, it was not the compiler version. I built in different modes (debug/release). I removed the code differences between debug/release mode.

bernhardmgruber:
Looks good! Please merge ;)

PR #43, opened 2020-07-20
Rename alpaka examples and delete Findpng++.cmake

bernhardmgruber:
I propose to rename the examples including alpaka, because I have a few examples, that I would like to add in subsequent PRs, that solve the same problem without alpaka. E.g. an nbody example without alpaka.


bertwesarg:
Why not using a new `alpaka` sub-directory?

I intent to install the `Alpaka*` headers from `examples/common` too, as I consider them of general use. I would put them also into an `alpaka` sub-directory, i.e., `include/llama/alpaka`.

bernhardmgruber:
The `alpaka` directory is a good idea! I did that!

As for alpaka support headers from the examples, I do not want to make a decision. I think Llama made a wrong decision here to concern itself with memory allocation as well. I think it would be better if you could just plug a Llama view on top of a user provided chunk of memory. Then we do not need to deal with allocators anymore. However, so far this is just an idea and I also do not have much time to implement it for now.

PR #44, opened 2020-07-21
Move Alpaka support headers into public include

bertwesarg:


bernhardmgruber:
I just had a VC with @psychocoderHPC and we discussed this PR among other issues. In principle, we are in favor of moving the alpaka support headers into llama itself. However, we would like to see a clear distinction between LLAMA only code and LLAMA code that depends on alpaka.

Therefore, we want to ask you to move the alpaka support headers in a dedicated subfolder, like `include/alpaka`. Furthermore, we would like to have a target `llama::alpaka` distinct from `llama::llama`, that has an include directory set to `include/alpaka`. This way, a LLAMA user can link against `llama::llama` for the core functionality (and not see the alpaka support headers) and against `llama::llama` and `llama::alpaka` to use LLAMA with alpaka support.

Could you please apply these changes and then rebase again onto develop? Thank you very much!

bertwesarg:
> Therefore, we want to ask you to move the alpaka support headers in a dedicated subfolder, like `include/alpaka`.

Its not clear to me, where the headers should be installed. Currently I install one into `<include>/llama/allocator/Alpaka.hpp`, because this is implements the LLAMA allocator interface. The other two go into `<include>/llama/alpaka`.

In case of your suggestion, I could put the allocator into

* `<include>/alpaka/llama/allocator/Alpaka.hpp`

and the other two into:

* `<include>/alpaka/llama/`

Is this what you two head in mind? I'm a little bit reluctant, to install files into another projects include dir, and thus optioned for `llama/alpaka` instead.

> Furthermore, we would like to have a target `llama::alpaka` distinct from `llama::llama`, that has an include directory set to `include/alpaka`. This way, a LLAMA user can link against `llama::llama` for the core functionality (and not see the alpaka support headers) and against `llama::llama` and `llama::alpaka` to use LLAMA with alpaka support.

This sounds like a good idea. Will try.

bernhardmgruber:
> I'm a little bit reluctant, to install files into another projects include dir, and thus optioned for `llama/alpaka` instead.

You are absolutely right! We should not install llama headers, even if they implement alpaka support, into `include/alpaka`. I guess the only option then remains to leave the alpaka support headers within `include/llama`, where they are reachable after linking against the `llama::llama` target.

If you can come up with something clever around a separate `llama::alpaka` target, that makes the alpaka support headers reachable, distinct from the headers of `llama::llama`, feel free to implement that! Otherwise, I think I can accept this PR without an additional `llama::alpaka` target.

@psychocoderHPC do you have an idea on where we could place/install the alpaka support headers? `include/llama/alpaka` makes them available right away, `include/alpaka` might install into an existing alpaka installation, `include/llama_alpaka` looks bad I think, but would probably allow a separate target. What do you think?

bertwesarg:
Even the `llama` headers are right away accessible, because we export `$CMAKE_INSTALL_PREFIX/include` as the include dir, thus the users is forced to use `#include <llama/llama.hpp>` to access the headers. Therefore, in does not matter in which directory we install the alpaka support headers, they will always be accessible to the user.

bernhardmgruber:
I pinged @psychocoderHPC again and he is fine with merging the PR as is.

@bertwesarg could you please rebase again to our current develop branch? Thank you!

bernhardmgruber:
So a bit of time has passed and I refactored away the Alpaka support except `ThreadsElemsDistribution` (which I also do not like a lot). One can now create `llama::View`s without allocators and just wrap them over an Alpaka buffer, which is much easier to use and also more cleanly draws the line between LLAMA and Alpaka.

I consider this PR therefore obsolete. Still, thank you for submitting it and the discussion! I am certain that I will need to create some Alpaka support again in the future. So the `llama::alpaka` CMake target and the filesystem organization will be relevant then.

PR #93, opened 2020-09-01
Refactoring

bernhardmgruber:


bussmann:
Hi, I see these refactoring things so offen. More comments and verbosity are highly welcome

bernhardmgruber:
If this was a production ready library, I would certainly follow a more rigorous workflow. However, I consider llama as a playground for the time being so I do not bother too much. The commit messages usually look a bit better though.

What's weird though with my current workflow are all the PRs. I should honestly just work on `develop` straight. But I kind of like the two level review. First the individual commit. And then a second look at the bigger picture when I create a PR and merge a full feature or a refactoring.

I am also surprised you are notified of these PRs. That must be a lot of emails! :D

PR #111, opened 2020-10-12
Enable multiple blobs for SoA mapping

bernhardmgruber:
A multi-blob version of the SoA mapping allows one blob per datum domain attribute.

### DIFF
--- a/examples/alpaka/nbody/nbody.cpp
+++ b/examples/alpaka/nbody/nbody.cpp
@@ -16,7 +16,7 @@
 #include <random>
 #include <utility>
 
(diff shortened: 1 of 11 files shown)


bussmann:
What is a blob?

psychocoderHPC:
image
Called: blob fish :smile:

bussmann:
Now I want multiple of these! Make it happen!

bernhardmgruber:
> What is a blob?

I simple search engine query could have answered that: A blob is a Binary large object.
Blobs are also described in the LLAMA documentation: https://llama-doc.readthedocs.io/en/latest/pages/blobs.html

Please also refrain from posting pictures of depressurized animals, even though the might look funny to someone ;)

PR #128, opened 2020-11-17
Vc alpaka nbody

bernhardmgruber:
Redesign the alpaka nbody to use Vc inside the kernel. The goal is to have a single source nbody version that performs optimal on GPU and CPU.

### DIFF
--- a/examples/alpaka/nbody/CMakeLists.txt
+++ b/examples/alpaka/nbody/CMakeLists.txt
@@ -1,11 +1,25 @@
 cmake_minimum_required (VERSION 3.15)
 project(llama-alpaka-nbody)
 
+find_package(Vc REQUIRED)
 if (NOT TARGET llama::llama)
 	find_package(llama REQUIRED)
 endif()
 find_package(alpaka 0.5.0 REQUIRED)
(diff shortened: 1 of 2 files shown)


examples/alpaka/nbody_vc/nbody.cpp
```diff
@@ -0,0 +1,324 @@
...
+#if defined(ALPAKA_ACC_CPU_B_SEQ_T_SEQ_ENABLED) || defined(ALPAKA_ACC_CPU_B_OMP2_T_SEQ_ENABLED)
+static constexpr std::size_t elements = Vc::float_v::size();
+static constexpr std::size_t threadsPerBlock = 1;
+constexpr auto aosoaLanes = elements; // vectors
+#elif defined(ALPAKA_ACC_GPU_CUDA_ENABLED)
+static constexpr std::size_t elements = 1;
+static constexpr std::size_t threadsPerBlock = 256;
+constexpr auto aosoaLanes = 32; // coalesced memory access
+#else
+#    error "Unsuppored backend"
```
ax3l:
```suggestion
#    error "Unsupported backend"
```

bernhardmgruber: Thanks for spotting that!

PR #129, opened 2020-11-19
CUDA nbody example and benchmark

bernhardmgruber:


bernhardmgruber:
Here are some numbers from an NVIDIA titanX:
image

GM = global memory, SM = shared memory, AoS = Array of Structs, SoA = Struct of Arrays, AoSoA32 = Array of Structs of 32 element Arrays, Y axis = update kernel runtime in seconds

examples/cuda/nbody/CMakeLists.txt
```diff
@@ -0,0 +1,12 @@
+cmake_minimum_required (VERSION 3.18.3)
+project(llama-cuda-nbody CXX CUDA)
+set(CMAKE_CUDA_ARCHITECTURES "35;75" CACHE STRING "CUDA architectures to compile for")
```
bernhardmgruber: What is the correct way to specify the CUDA_ARCHITECTURES to build for? I have read, that you should not specify them in the build script, because users should specify this themselves. But for easy of use, I would like to provide a default, so one can also omit it and still have a compiling example.

j-stephan: What about

```cmake
if(NOT DEFINED CMAKE_CUDA_ARCHITECTURES)
    set(CMAKE_CUDA_ARCHITECTURES ...)
endif()
```

Or am I missing something here?

bernhardmgruber: @j-stephan what values should I place where your `...` are? Which architectures by default? And should I provide a default **at all**?

psychocoderHPC: We set by default the minimum version supported by the used CUDA library. The user then should define the used version explicitly.
We set the command in all our projects that we embed the binary GPU code + the ptx code. This has the advantage that code compiled for e.g. sm_35 can be jit compiled to all newer architecture. Jit compiling has the disadvantage that not all features of the architecture will be used and the start of the application could take very long. Therefore to set the architecture explicit is always recommended.

psychocoderHPC: If you use native CUDA support in CMake it could be that you need to set a special flag to embed the ptx code. I never tested native CUDA suppot.

j-stephan: > what values should I place where your ... are? Which architectures by default?

If you don't like the CMake defaults - whatever you believe to be sensible ;-) IIRC we don't supply a default in alpaka.

> And should I provide a default at all?

It depends. Do you have issues with the [CMake defaults](https://cmake.org/cmake/help/v3.18/variable/CMAKE_CUDA_ARCHITECTURES.html)?

psychocoderHPC: We set the default in alpaka: https://github.com/alpaka-group/alpaka/blob/d67c7f64a8cdc555ab6977eeffcc0b703f98f9ee/cmake/alpakaCommon.cmake#L288-L292

bernhardmgruber: @psychocoderHPC thank you for the feedback! This is what I was looking for. Therefor with
```
set(CMAKE_CUDA_ARCHITECTURES "35" CACHE STRING "CUDA architectures to compile for")
```
I can set a minimum version. And users can override this inside a cmake GUI or by passing e.g. `-DCMAKE_CUDA_ARCHITECTURES=70`. I just tried that, it works.

psychocoderHPC: > Here are some numbers from an NVIDIA titanX:
> image

What is `GM` and `SM`?

psychocoderHPC:
> Here are some numbers from an NVIDIA titanX:
> image

What is `GM` and `SM`? What is shown on the Y axis?

bernhardmgruber:
GM = global memory, SM = shared memory, AoS = Array of Structs, SoA = Struct of Arrays, AoSoA32 = Array of Structs of 32 element Arrays, Y axis = update kernel runtime in seconds

PR #132, opened 2020-12-09
Various improvements to CUDA example

bernhardmgruber:


include/llama/View.hpp
```diff
@@ -21,27 +21,19 @@ namespace llama
...
+    /// allocator::Vector). The allocator callable is called with the size of bytes to allocate for each blob of the
+    /// mapping. This function is the preferred way to create a \ref View.
     template <typename Mapping, typename Allocator = allocator::Vector<>>
     LLAMA_FN_HOST_ACC_INLINE auto allocView(Mapping mapping = {}, const Allocator& alloc = {})
         -> View<Mapping, internal::AllocatorBlobType<Allocator>>
     {
-        auto blobs = internal::makeBlobArray<Allocator>(alloc, mapping, std::make_index_sequence<Mapping::blobCount>{});
+        Array<internal::AllocatorBlobType<Allocator>, Mapping::blobCount> blobs;
+        for (int i = 0; i < Mapping::blobCount; i++)
+            blobs[i] = alloc(mapping.getBlobSize(i));
```
bernhardmgruber: That might be a pessimization if the allocator's blob type is not cheap to construct.

PR #135, opened 2020-12-15
Tiled Vc version

bernhardmgruber:


examples/nbody/nbody.cpp
```diff
@@ -569,7 +569,8 @@ namespace manualAoSoA
     void updateTiled(ParticleBlock<Lanes>* particles)
     {
         constexpr auto blocks = PROBLEM_SIZE / Lanes;
-        constexpr auto blocksPerTile = 64; // L1D_SIZE / sizeof(ParticleBlock<Lanes>);
+        constexpr auto blocksPerTile = 128; // L1D_SIZE / sizeof(ParticleBlock<Lanes>);
```
bernhardmgruber, line 572: Optimally, we could use `L1D_SIZE / sizeof(ParticleBlock<Lanes>)`, but that does not evenly divide the problem size and would require some more code around. So I took the next lower power of 2.

PR #153, opened 2021-02-08
Remove CUDA installer cache and update to CUDA 11.2

bernhardmgruber:


bernhardmgruber:
It turns out that uncaching takes 2min 30s because of the untaring of the archive. A fresh download takes 30s. So I disabled the CUDA installer cache again.

PR #155, opened 2021-02-25
Split mapping improvements

bernhardmgruber:
* Some fixes with incorrectly calculated blob sizes
* Blob splitting should work now
* Split mappings can be nested
* renamed from SplitMapping to Split
* Added some more tests


bernhardmgruber:
Here is an interesting split of 32 particles:
```c++
llama::mapping::Split<
            ArrayDomain,
            Particle,
            llama::DatumCoord<1>,
            llama::mapping::PreconfiguredAoSoA<8>::type,
            llama::mapping::PreconfiguredSplit<
                llama::DatumCoord<1>,
                llama::mapping::One,
                llama::mapping::
                    PreconfiguredSplit<llama::DatumCoord<0>, llama::mapping::AoS, llama::mapping::SoA, true>::type,
                true>::type,
            true>
```

image

PR #156, opened 2021-02-25
Extend sizeOf<> and offsetOf<> to support alignment and padding

bernhardmgruber:
If we turn this on in the AoS mapping, we can turn this mess:
image
into this nicety:
image

Afterwards we would need a struct member permutation algorithm that minimizes padding. But that's for another day.

### DIFF
--- a/include/llama/Core.hpp
+++ b/include/llama/Core.hpp
@@ -272,14 +272,14 @@ namespace llama
     namespace internal
     {
         template <typename T, std::size_t... Coords, typename Functor>
-        LLAMA_FN_HOST_ACC_INLINE constexpr void forEachLeaveImpl(T, DatumCoord<Coords...> coord, Functor&& functor)
+        LLAMA_FN_HOST_ACC_INLINE constexpr void forEachLeaveImpl(T*, DatumCoord<Coords...> coord, Functor&& functor)
         {
             functor(coord);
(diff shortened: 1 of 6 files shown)


include/llama/Core.hpp
```diff
@@ -396,6 +357,65 @@ namespace llama
...
+    template <typename DatumDomain, typename DatumCoord, bool Align = false>
+    inline constexpr std::size_t offsetOf = []() constexpr
+    {
+        if constexpr (Align)
+        {
+            constexpr auto i = flatDatumCoord<DatumDomain, DatumCoord>;
+            using Tuple = boost::mp11::mp_rename<boost::mp11::mp_reverse<FlattenDatumDomain<DatumDomain>>, std::tuple>;
+            constexpr auto n = boost::mp11::mp_size<Tuple>::value;
+            Tuple t;
+            return (std::intptr_t) std::addressof(std::get<n - 1 - i>(t)) - (std::intptr_t) std::addressof(t);
```
bernhardmgruber: It turns out this is accepted by g++ 10.2, but not by g++ trunk (upcoming g++ 11.0) or MSVC.

bernhardmgruber: Replaced by custom implementation not relying on `std::tuple`.

PR #158, opened 2021-03-03
Add Iterator to make the STL available to 1D Views

bernhardmgruber:


tests/iterator.cpp
```diff
@@ -0,0 +1,60 @@
...
+    auto soaView = llama::allocView(llama::mapping::SoA<ArrayDomain, Position>{arrayDomain});
+
+    int i = 0;
+    for (auto vd : aosView)
+    {
+        vd(tag::X{}) = ++i;
+        vd(tag::Y{}) = ++i;
+        vd(tag::Z{}) = ++i;
+    }
+    std::copy(begin(aosView), end(aosView), begin(soaView));
```
bernhardmgruber, line 52: Having this line work is super crazy! This means we can use `std::copy` to copy between ANY LLAMA layouts. This will always do an element-wise copy, which is sometimes suboptimal. But it correctly shuffles the layouts through the iterators and virtual datums.

PR #161, opened 2021-03-04
Remove unused header and replace boost::replace_all

bernhardmgruber:
* Remove unused header
* Replace boost::replace_all by custom implementation. This avoids pulling in boost::mpl, which causes issues with nvcc, and improves compile times.


bussmann:
Could you choose a more verbose title for this, please? This won't be the last refactoring, I suppose.

bernhardmgruber:
Does the new title sound better to you? I will try to word them better in the future.

bussmann:
Much better, thank you!

PR #165, opened 2021-03-08
Simplify offsetof and sizeof

bernhardmgruber:
* simplify offsetof
* simplify sizeof
* workaround MSVC bug

### DIFF
--- a/include/llama/Core.hpp
+++ b/include/llama/Core.hpp
@@ -401,53 +401,27 @@ namespace llama
     }
     ();
 
-    namespace internal
-    {
-        template <bool Align, typename T>
-        constexpr auto offsetOfImpl(T*, DatumCoord<>)
-        {
-            return 0;
-        }
-
-        template <bool Align, typename... DatumElements, std::size_t FirstCoord, std::size_t... Coords>
-        constexpr auto offsetOfImpl(DatumStruct<DatumElements...>*, DatumCoord<FirstCoord, Coords...>)
-        {
-            std::size_t acc = 0;
-            boost::mp11::mp_for_each<boost::mp11::mp_take_c<DatumStruct<DatumElements...>, FirstCoord>>([&](
-                auto element) constexpr { acc += sizeOf<GetDatumElementType<decltype(element)>>; });
-
-            using Element = boost::mp11::mp_at_c<DatumStruct<DatumElements...>, FirstCoord>;
-            acc += offsetOfImpl<Align>((GetDatumElementType<Element>*) nullptr, DatumCoord<Coords...>{});
-
-            return acc;
-        }
-    } // namespace internal
-
-    /// The byte offset of an element in a datum domain if it would be a normal
-    /// struct.
+    /// The byte offset of an element in a datum domain if it would be a normal struct.
     /// \tparam DatumDomain Datum domain tree.
     /// \tparam DatumCoord Datum coordinate of an element indatum domain tree.
     template <typename DatumDomain, typename DatumCoord, bool Align = false>
     inline constexpr std::size_t offsetOf = []() constexpr
     {
-        if constexpr (Align)
-        {
-            using namespace boost::mp11;
+        using namespace boost::mp11;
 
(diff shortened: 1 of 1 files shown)


bussmann:
Please consider meaningful descriptions of your commits. Later readers will thank you.

bernhardmgruber:
> Please consider meaningful descriptions of your commits. Later readers will thank you.

Then please help me how

* simplify offsetof
* simplify sizeof
* workaround MSVC bug

could be written more meaningful without getting unnecessary verbose for a non-functional change. I think the description perfectly says what these changes are. Just a small simplification and a workaround for MSVC. There is no bugfix and no new behavior. It's just a simplification.

ax3l:
We usually write a PR description with an easy-to-grasp summary.

Simplify offsetof/sizeof can mention that it simplifies the internal implementation (aka "does not change the user-facing API") and potentially a half-sentence/keyword on the used technique (e.g. "modernize with C++17 `if constexpr`"). This is helpful when writing changelogs and revisiting PRs for whatever reason.

`workaround MSVC bug` could quote the error message for google-ability and document the MSVC version that triggers the need for a work-around. That way, one can find the right PR if one faces the error (e.g. using an older version) and people are able to remove a work-around later on if a compiler version is dropped from supported targets.

bussmann:
Thank you, @ax3l !

bernhardmgruber:
> We usually write a PR description with an easy-to-grasp summary.

In this case I really believe the PR's title is enough. Why repeat it?

> `workaround MSVC bug` could quote the error message for google-ability and document the MSVC version that triggers the need for a work-around. That way, one can find the right PR if one faces the error (e.g. using an older version) and people are able to remove a work-around later on if a compiler version is dropped from supported targets.

Does that really help? If you want to reproduce the bug, you can just checkout the commit before the workaround. And the error novel is many pages long. I would not want to paste that into the commit log. Furthermore, it is not even clear what the exact error is. I tried a few variants and either got an internal compile error or an error that I had a read from an uninitialized variable inside a constexpr function. clang and gcc always compiled fine. I am sure the previous code was correct, so I concluded it must be an issue with MSVC.

The failing MSVC version is the most current one (1929), but there is no list of supported compilers for LLAMA yet. So the workaround is just marked with a TODO and should be removed as soon as there is a fix in MSVC. And this is fine because LLAMA is not yet a shipping product. It is a research prototype. Everything is volatile and changing. I am totally fine with putting in compiler version numbers at some point and specifying a list of supported environments, platforms, compilers etc., when LLAMA has its 1.0 release or even earlier when we have an actual user base.

> Simplify offsetof/sizeof can mention that it simplifies the internal implementation (aka "does not change the user-facing API") and potentially a half-sentence/keyword on the used technique (e.g. "modernize with C++17 if constexpr"). This is helpful when writing changelogs and revisiting PRs for whatever reason.

I could prefix a `[NFC]` next time, that is right! Thx!

I would not know how to describe the used technique. The C++17 `if constexpr` was there before. What I did was collapse this:
```c++
if constexpr Align
    // implementation A
else
    // implementation B
```
into
```c++
// unified implementation
if constexpr Align
    // detail of previous implementation A
// unified implementation
```
How would you describe such a transformation? Is this a known refactoring pattern? I could think of "collapsed aligned and unaligned offsetof implementation into a unified one". But does that really help? With this description I still need to consult the diff if I really want to know what happended. So I think it does not add any value. Besides, such a description is more wordy and suggests a more profound change to me. But this could just be my intuition.

Lastly, I would even prefer not to create such PRs at all. I would be most happy to just rebase my own branches and then push them to `develop`. However, I got bitten once or twice by having the CI fail afterwards, since I test different compilers locally (gcc10 + MSVC) before committing. So I find a PR is a usefull tool to reflect again on the changeset I want to merge and to ensure that a few compilers can really build the stuff. That's my main motivation for such PRs.
For actual new features or examples, I usually create far richer PRs that showcase some of the new functionality and give motivation: https://github.com/alpaka-group/llama/pull/149

PR #171, opened 2021-03-12
Support computed properties

### LINKED ISSUE #170: Prototype computed fields
There are cases when defining a data structure with a set of elements, where some elements are functionally dependent on other elements. A good example is a triangle storing its 3 vertices and its normal vector. In some cases it makes sense to store the normal as part of a triangle object because computing it is expensive (2 vector subtractions, 1 cross product, 1 normalization). There might be cases though, where we value the memory footprint more than the computational costs. Or their might be architectures where the recomputation is still faster then fetching additional data from memory.

LLAMA should allow mappings where the mapping can decide to not store certain elements, but compute them on access.

This allows further interesting and edgy use cases like fully computed LLAMA views, having no memory footprint anymore.

bernhardmgruber:
Add a hook to View, testing whether a mapping exposes a member function `isComputed(DatumCoord)` and if the value is true for a given DatumCoord on access. If true, the `compute()` member function of the mapping is invoked instead of `getBlobNrAndOffset()`, allowing the mapping to provide a computed value instead of a memory location.

### DIFF
--- a/include/llama/View.hpp
+++ b/include/llama/View.hpp
@@ -844,6 +844,20 @@ namespace llama
         View* view;
     };
 
+    namespace internal
+    {
+        template <typename Mapping, typename DatumCoord, typename = void>
+        struct isComputed : std::false_type
+        {
+        };
+
+        template <typename Mapping, typename DatumCoord>
+        struct isComputed<Mapping, DatumCoord, std::void_t<decltype(Mapping::isComputed(DatumCoord{}))>>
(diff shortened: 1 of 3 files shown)


bernhardmgruber:
Although this PR is small it opens up a whole new world of mappings. I consider releasing a new LLAMA version before we open up this possibilities.

bernhardmgruber:
Computed properties are already needed for some new LLAMA demonstrations so I include the feature now.

PR #174, opened 2021-03-14
[NFC] Add a comment why only 1D iterators are supported

bernhardmgruber:


bussmann:
One could always map a higher dimensional iterator to 1D.

bernhardmgruber:
Sure, LLAMA already does that. The problem is we would need to map a 1D iterator into an ND `ArrayDomain` coordinate, which is later relinearized by the mapping. I have not yet found a way to avoid remapping twice. And the cost of these remappings outweights any penality from bad memory access patterns.

PR #224, opened 2021-05-03
Add SoA <-> AoSoA viewcopy and improve plot

bernhardmgruber:
* plot throughput instead of runtime
* extend `aosoa_copy` to also support SoA
* repeat copy a few times to reduce noise
* fix using the wrong OpenMP API to retrieve the thread count
* add title to plot with hostname and total copied size

### DIFF
--- a/examples/viewcopy/viewcopy.cpp
+++ b/examples/viewcopy/viewcopy.cpp
@@ -380,7 +380,7 @@ inline constexpr auto is_AoSoA<llama::mapping::AoSoA<AD, RD, L>> = true;
 auto main() -> int
 try
 {
-    const auto numThreads = static_cast<std::size_t>(omp_get_num_threads());
+    const auto numThreads = static_cast<std::size_t>(omp_get_max_threads());
     std::cout << "Threads: " << numThreads << "\n";
 
     const auto arrayDims = llama::ArrayDims{1024, 1024, 16};
--- a/examples/viewcopy/viewcopy.cpp
+++ b/examples/viewcopy/viewcopy.cpp
@@ -10,6 +10,8 @@
 #include <omp.h>
 #include <string_view>
 
+constexpr auto REPETITIONS = 5;
+
 // clang-format off
 namespace tag
 {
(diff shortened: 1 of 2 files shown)


bernhardmgruber:
image

The results are quite surprising, because the parallel `naive_copy` is performing far better than expected. It also has zero LLC misses, so the hardware seems to understand it pretty well.
The parallel `aosoa_copy` versions are pretty strong, as expected.
But what is really surprising is that parallel `naive_copy` and `aosoa_copy` both outperform a parallel `memcpy` in several benchmarks. So we added a hand-rolled `memcopy_G` from @amadio which is tuned for larger block sizes.

examples/viewcopy/viewcopy.cpp
```diff
@@ -96,6 +102,62 @@ void std_copy(
     std::copy(srcView.begin(), srcView.end(), dstView.begin());
 }
 
+// see: https://stackoverflow.com/posts/30386256/revisions
```
amadio:
```suggestion
// see: https://stackoverflow.com/a/30386256/1034717
```

examples/viewcopy/viewcopy.cpp
```diff
@@ -96,6 +102,62 @@ void std_copy(
...
+void* memcpy_Guilherme(void* dest, const void* src, size_t n)
```
amadio: Maybe just call this `memcpy_avx`...

bernhardmgruber: Ok!

examples/viewcopy/viewcopy.cpp
bernhardmgruber: Thx!

examples/viewcopy/viewcopy.cpp
```diff
@@ -96,7 +102,59 @@ void std_copy(
...
+        assert(header_bytes < 32);
+
+        memcpy(d, s, std::min(header_bytes, n));
+
+        d = reinterpret_cast<char*>(ALIGN(reinterpret_cast<uintptr_t>(d), 32));
+        s = reinterpret_cast<char*>(ALIGN(reinterpret_cast<uintptr_t>(s), 32));
+        n -= std::min(header_bytes, n);
+    }
+
+    constexpr auto unrollFactor = 8;
```
bernhardmgruber, line 130: I did some experiments with the unroll factor:
with unroll 16:
```
memcpy_avx2 1.18689 s   7.37223GiB/s  
memcpy_avx2(p) 0.737673 s       11.8616GiB/s  
```
with unroll 8:
```
memcpy_avx2 1.18048 s   7.41226GiB/s      
memcpy_avx2(p) 0.701376 s       12.4755GiB/s   
```
with unroll 4:
```
memcpy_avx2 1.18842 s   7.36272GiB/s     
memcpy_avx2(p) 0.762377 s       11.4773GiB/s   
```
with unroll 2:
```
memcpy_avx2 1.21957 s   7.17467GiB/s
memcpy_avx2(p) 0.748464 s       11.6906GiB/s
```
handwritten version equivalent to unroll 2:
```
memcpy_avx2 1.20118 s   7.28448GiB/s 
memcpy_avx2(p) 0.736476 s       11.8809GiB/s        
```
It seems an unroll factor of 8 works quite ok. But all of this is probably very architecture specific.

Issue #240, opened 2021-05-16
Improve compile time


bernhardmgruber:
Many LLAMA constructs depend on compile time iteration of type lists. For very big record dimensions, some constructs become painfully slow.

Consider the following clang timetrace from a LLAMA example using the ROOT HEP framework:
![image](https://user-images.githubusercontent.com/1224051/118399157-1b86d600-b65c-11eb-8b25-692a9eeea5f8.png)
The first part marked with "ROOT" is parsing ROOT's headers, the part marked with "D" is LLAMA's dumping code. The part between the black bars is parsing LLAMA's headers (~30ms). Then comes a very long part "InstantiateFunction", which comes from this code:
```c++
llama::forEachLeaf<Event>([&](auto coord) {
    using Name = llama::GetTag<Event, decltype(coord)>;
    using Type = llama::GetType<Event, decltype(coord)>;
    auto column = ntuple->GetView<Type>(llama::structName<Name>());
    for (std::size_t i = 0; i < n; i++)
        view(i)(coord) = column(i);
});
```
The big issue here is that the `Event` record dimension is large (~200 fields). The loop created by `forEachLeaf` is linear in compilation time and not yet the problem, but `view(i)(coord)` eventually calles into the `AoS::blobNrAndOffset`, which calles `offsetOf<RD, RC>`, and the implementation of `offsetOf` is linear again. Thus, the whole code snipped is quadratic in compilation time.

The situation could be improved a lot by making use of template instantiation memoization. That is, if the value of `offsetOf<RD, RC>` could somehow depend on the value of `offsetOf<RD, RC - 1>`, the compiler could reuse memoized previous template instantiations and the code snipped should compile in linear time.

The big difficulty is that record coordinates are hierarchical and `RC - 1` is a complex operation. A good solution would be to linearize record coordinates before they are passed to mappings, and mappings deal solely in linear coordinates.

bernhardmgruber:
#241 improved the situation a lot. However, for a `llama::Record` with over 1000 fields, clang still takes around 2min to compile. This is a lot better, but still annoying.

PR #245, opened 2021-05-17
[on hold] RNTuple example

bernhardmgruber:
This is an example for using LLAMA with data loaded from an RNTuple via the ROOT data analysis framework.

At the moment, a LLAMA view is created to hold all events stored in the RNTuple file, excluding subarrays. This data contains 1095 columns.

It compiles in 2min, needs ~2GiB of memory at runtime and copies from RNTuple to LLAMA in 26s.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -86,6 +86,12 @@ if (LLAMA_BUILD_EXAMPLES)
 		message(WARNING "Could not find alpaka. Alpaka examples are disabled.")
 	endif()
 
+	# ROOT examples
+	find_package(ROOT QUIET)
+	if (ROOT_FOUND)
+		add_subdirectory("examples/hep_rntuple")
+	endif()
+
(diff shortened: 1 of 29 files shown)


bussmann:
Can you put these numbers in perspective?

bernhardmgruber:
> Can you put these numbers in perspective?

As a quick answer, I can only provide you with experience and guesses:
* 2min compilation time is long for a single source file. Other LLAMA examples compile within ~10s. LLAMA has scalability issues with very large records, although I tried to improve the situation a lot in the recent days with 2 PRs against #240.
* The RNTuple on disk (compressed) has 2.5GiB. I read the events, which are roughly 2/3 of the data. If that is 2GiB uncompressed in RAM, that sounds reasonable. I guess the compression is not that strong.
* Reading that data in 26s seems slow, but it runs in WSL on my notebook. And I have not made any attempt to improve that.

PR #289, opened 2021-06-04
Lift copy implementations into llama

### LINKED ISSUE #272: Move view copy into LLAMA
A generic copy routine between views has been developed as part of the viewcopy example. We should move this implementation into the LLAMA library itself.

bernhardmgruber:
The viewcopy example prototyped several implementations for a mapping aware copy between two views. This PR lifts these into LLAMA.

- [x] verify they perform the same as before being lifted into LLAMA
- [x] add unit tests
- [x] add documentation

### DIFF
--- a/docs/pages/copying.rst
+++ b/docs/pages/copying.rst
@@ -16,10 +16,6 @@ E.g. when both mappings use SoA, but one time with, one time without padding, or
 Or two AoSoA mappings with a different inner array length.
 In those cases an optimized copy procedure is possible, copying larger chunks than mere fields.
 
-.. For the moment, LLAMA implements a generic, field-wise copy with specializations for combinations of SoA and AoSoA mappings, reflect the properties of these.
-.. This is sub-optimal, because for every new mapping new specializations are needed.
-
-.. One thus needs new approaches on how to improve copying because LLAMA can provide the necessary infrastructure:
 Four solutions exist for this problem:
 
 1. Implement specializations for specific combinations of mappings, which reflect the properties of these.
@@ -42,3 +38,42 @@ A good approach could use smaller intermediate views to shuffle a chunk from one
 The `async copy example <https://github.com/alpaka-group/llama/blob/master/examples/asynccopy/asynccopy.cpp>`_ tries to show an asynchronous copy/shuffle/compute workflow.
 This example applies a bluring kernel to an RGB-image, but also may work only on two or one channel instead of all three.
 Not used channels are not allocated and especially not copied.
+
+
+For the moment, LLAMA implements a generic, field-wise copy algorithm with faster specializations for combinations of SoA and AoSoA mappings.
+
+.. code-block:: C++
+
+    auto srcView = llama::allocView(srcMapping);
+    auto dstView = llama::allocView(dstMapping);
+    llama::copy(srcView, dstView); // use best copy strategy
+
+Internally, :cpp:`llama::copy` will choose a copy strategy depending on the source and destination mapping.
+This choice is done via template specializations of the :cpp:`llama::Copy` class template.
+Users can add specializations of :cpp:`llama::Copy` to provide additional copy strategies:
+
+.. code-block:: C++
+
+    // provide special copy from AoS -> UserDefinedMapping
+    template <typename ArrayDims, typename RecordDim, bool Aligned, typename LinearizeArrayDims>
+    struct Copy<
+        llama::mapping::AoS<ArrayDims, RecordDim, Aligned, LinearizeArrayDims>,
+        UserDefinedMapping<ArrayDims, RecordDim>>
+    {
+        template <typename SrcBlob, typename DstBlob>
+        void operator()(
+            const View<mapping::AoS<ArrayDims, RecordDim, Aligned, LinearizeArrayDims>, SrcBlob>& srcView,
+            View<mapping::SoA<ArrayDims, RecordDim, DstSeparateBuffers, LinearizeArrayDims>, DstBlob>& dstView,
+            std::size_t threadId, std::size_t threadCount) {
+            ...
+        }
+    };
+
+    llama::copy(srcView, dstView); // can delegate to above specialization now
+
+LLAMA also allows direct access to its two copy implementations, which is mainly used for benchmarking them:
+
+.. code-block:: C++
+
+    llama::fieldWiseCopy(srcView, dstView); // explicit field-wise copy
+    llama::aosoaCommonBlockCopy(srcView, dstView); // explicit SoA/AoSoA copy
--- a/examples/viewcopy/viewcopy.cpp
+++ b/examples/viewcopy/viewcopy.cpp
@@ -46,48 +46,6 @@ using Particle = llama::Record<
 using RecordDim = boost::mp11::mp_take_c<Event, 20>;
 // using RecordDim = Event; // WARN: expect long compilation time
 
-namespace llamaex
-{
-    using namespace llama;
-
-    template <std::size_t Dim, typename Func>
-    void parallelForEachADCoord(ArrayDims<Dim> adSize, std::size_t numThreads, Func&& func)
-    {
-#pragma omp parallel for num_threads(numThreads)
-        for (std::ptrdiff_t i = 0; i < static_cast<std::ptrdiff_t>(adSize[0]); i++)
-        {
-            if constexpr (Dim > 1)
-                forEachADCoord(
-                    ArrayDims<Dim - 1>{pop_front(adSize)},
-                    std::forward<Func>(func),
-                    static_cast<std::size_t>(i));
-            else
-                std::forward<Func>(func)(ArrayDims<Dim>{static_cast<std::size_t>(i)});
-        }
-    }
-} // namespace llamaex
-
-template <typename SrcMapping, typename SrcBlobType, typename DstMapping, typename DstBlobType>
-void naive_copy(
-    const llama::View<SrcMapping, SrcBlobType>& srcView,
-    llama::View<DstMapping, DstBlobType>& dstView,
-    std::size_t numThreads = 1)
-{
-    static_assert(std::is_same_v<typename SrcMapping::RecordDim, typename DstMapping::RecordDim>);
-
-    if (srcView.mapping.arrayDims() != dstView.mapping.arrayDims())
-        throw std::runtime_error{"Array dimensions sizes are different"};
-
-    llamaex::parallelForEachADCoord(
-        srcView.mapping.arrayDims(),
-        numThreads,
-        [&](auto ad) LLAMA_LAMBDA_INLINE
-        {
-            llama::forEachLeaf<typename DstMapping::RecordDim>([&](auto coord) LLAMA_LAMBDA_INLINE
-                                                               { dstView(ad)(coord) = srcView(ad)(coord); });
-        });
-}
-
 template <typename SrcMapping, typename SrcBlobType, typename DstMapping, typename DstBlobType>
 void std_copy(const llama::View<SrcMapping, SrcBlobType>& srcView, llama::View<DstMapping, DstBlobType>& dstView)
 {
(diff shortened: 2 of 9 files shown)


docs/pages/copying.rst
```diff
@@ -42,3 +38,42 @@ A good approach could use smaller intermediate views to shuffle a chunk from one
 The `async copy example <https://github.com/alpaka-group/llama/blob/master/examples/asynccopy/asynccopy.cpp>`_ tries to show an asynchronous copy/shuffle/compute workflow.
 This example applies a bluring kernel to an RGB-image, but also may work only on two or one channel instead of all three.
 Not used channels are not allocated and especially not copied.
+
+
+For the moment, LLAMA implements a generic, field-wise copy with specializations for combinations of SoA and AoSoA mappings, reflect the properties of these.
```
j-stephan: This sentence sounds a bit weird.

docs/pages/copying.rst
```diff
@@ -42,3 +38,42 @@ A good approach could use smaller intermediate views to shuffle a chunk from one
...
+
+.. code-block:: C++
+
+    auto srcView = llama::allocView(srcMapping);
+    auto dstView = llama::allocView(dstMapping);
+    llama::copy(srcView, dstView); // use best copy strategy
+
+Internally, :cpp:`llama::copy` will choose a copy strategy depending on the source and destination mapping.
+This choice is done via template specializations of the :cpp:`llama::Copy` class template.
+Users can add specializations of :cpp:`llama::Copy` to provide additional copy stragegies:
```
j-stephan: Typo:

```suggestion
Users can add specializations of :cpp:`llama::Copy` to provide additional copy strategies:
```

examples/viewcopy/viewcopy.cpp
```diff
@@ -472,11 +191,19 @@ set ylabel "throughput [GiB/s]"
 #endif
     benchmarkMemcpy(
         "memcpy(p)",
-        [&](auto* dst, auto* src, auto size) { parallel_memcpy(dst, src, size, std::memcpy, numThreads); });
+        [&](auto* dst, auto* src, auto size)
+        {
+#pragma omp parallel
```
j-stephan, line 196: This will hopefully get solved with clang-format 12.

include/llama/mapping/AoS.hpp
```diff
@@ -10,17 +10,18 @@ namespace llama::mapping
     /// Array of struct mapping. Used to create a \ref View via \ref allocView.
     /// \tparam AlignAndPad If true, padding bytes are inserted to guarantee that struct members are properly aligned.
     /// If false, struct members are tighly packed.
-    /// \tparam LinearizeArrayDimsFunctor Defines how the array dimensions should be mapped into linear numbers and
+    /// \tparam T_LinearizeArrayDimsFunctor Defines how the array dimensions should be mapped into linear numbers and
```
j-stephan, line 13: I'm a bit confused by the naming here. Our guidelines mandate a `T` prefix (not `T_`). Did you change this yourself or is the clang-tidy file wrong?

include/llama/Copy.hpp
```diff
@@ -0,0 +1,364 @@
...
+{
+    namespace internal
+    {
+        inline void parallel_memcpy(
+            std::byte* dst,
+            const std::byte* src,
+            std::size_t size,
+            std::size_t threadId = 0,
+            std::size_t threadCount = 1,
+            decltype(std::memcpy) memcpy = std::memcpy)
```
j-stephan: Is the `decltype(std::memcpy)` laziness or does it have a deeper meaning?

include/llama/Copy.hpp
```diff
@@ -0,0 +1,364 @@
...
+            const auto sizePerThread = size / threadCount;
+            const auto sizeLastThread = sizePerThread + size % threadCount;
+            const auto sizeThisThread = threadId == threadCount - 1 ? sizeLastThread : sizePerThread;
+            memcpy(dst + threadId * sizePerThread, src + threadId * sizePerThread, sizeThisThread);
+        }
+    } // namespace internal
+
+    /// @brief Direct memcpy from source view blobs to destination view blobs. Both views need to have the same mappings
+    /// with the same array dimensions.
+    /// @param threadId Optional. Thread id in case of multi-threaded copy.
```
j-stephan:
```suggestion
    /// @param threadId (Optional) ID of calling thread. Useful for multi-threaded applications. 
```

Similar in next line.

include/llama/Copy.hpp
```diff
@@ -0,0 +1,364 @@
...
+            const std::byte* src,
+            std::size_t size,
+            std::size_t threadId = 0,
+            std::size_t threadCount = 1,
+            decltype(std::memcpy) memcpy = std::memcpy)
+        {
+            const auto sizePerThread = size / threadCount;
+            const auto sizeLastThread = sizePerThread + size % threadCount;
+            const auto sizeThisThread = threadId == threadCount - 1 ? sizeLastThread : sizePerThread;
+            memcpy(dst + threadId * sizePerThread, src + threadId * sizePerThread, sizeThisThread);
```
j-stephan: Isn't this a possible shadowing of C's `memcpy`? Or any other `memcpy` in the global namespace, for that matter.

include/llama/Copy.hpp
```diff
@@ -0,0 +1,364 @@
...
+                return &dstView.storageBlobs[0][0] + mapAoSoA(flatArrayIndex, coord, LanesDst);
+            else
+            {
+                const auto [blob, off] = mapSoA(flatArrayIndex, coord, MBDst);
+                return &dstView.storageBlobs[blob][off];
+            }
+        };
+
+        constexpr auto L = std::min(LanesSrc, LanesDst);
+        static_assert(!srcIsAoSoA || LanesSrc % L == 0);
```
j-stephan: Can this fail for user code? If so, this needs an error message.

include/llama/Copy.hpp
```diff
@@ -0,0 +1,364 @@
...
+                            for (std::size_t i = start; i < stop; i += L)
+                                copyLBlock(threadDst, i, coord);
+                        });
+                }
+            }
+        }
+    }
+
+    /// @brief Generic implementation of \ref copy. LLAMA provides several specializations of this construct for
+    /// specific mappings. Users are encourages to also specialize this template with better copy algorithms.
```
j-stephan: Typo. Also:

```suggestion
    /// specific mappings. Users are encouraged to also specialize this template with better copy algorithms if they fit their use case.
```

j-stephan (changes requested):

examples/viewcopy/viewcopy.cpp
bernhardmgruber: LLAMA already uses clang-format 12 for several weeks :D But what do you think needs to be solved here?

include/llama/mapping/AoS.hpp
bernhardmgruber: I have not yet applied the full clang-tidy from CRP. Especially the renamings are missing. So this is consistent with how I named some of the old template parameters.

examples/viewcopy/viewcopy.cpp
j-stephan: The `#pragma` should be indented.

include/llama/Copy.hpp
bernhardmgruber: `decltype(std::memcpy)` is a shorter way to spell the function pointer type for `memcpy` :) This parameter is intended to provide a custom implementation for `memcpy`. E.g. the AVX2 `memcpy` in the viewcopy example.

include/llama/Copy.hpp
bernhardmgruber: Technically, C++ compilers should not provide `memcpy` in the global namespace. Practically, they often do. So yes, this is a shadowing. I will rename it.

include/llama/Copy.hpp
bernhardmgruber: Yes it can and you are right!

docs/pages/copying.rst
bernhardmgruber: Thx!

include/llama/Copy.hpp
bernhardmgruber: Thx, reworded!

docs/pages/copying.rst
bernhardmgruber: fixed

examples/viewcopy/viewcopy.cpp
bernhardmgruber: Ah, I see! Well that is for sure something clang-format can handle. But I am using the CRP `.clang-format`, so I guess the coding guidelines don't indent it.

include/llama/Copy.hpp
bernhardmgruber: Renamed parameter.

include/llama/Copy.hpp
bernhardmgruber: I replaced the code by a different implementation which does not have this problem.

include/llama/Copy.hpp
bernhardmgruber: Reworded.

include/llama/Copy.hpp
```diff
@@ -0,0 +1,381 @@
...
+    /// @param threadCount Optional. Thread count in case of multi-threaded copy.
+    template <typename SrcMapping, typename SrcBlob, typename DstMapping, typename DstBlob>
+    void fieldWiseCopy(
+        const View<SrcMapping, SrcBlob>& srcView,
+        View<DstMapping, DstBlob>& dstView,
+        std::size_t threadId = 0,
+        std::size_t threadCount = 1)
+    {
+        // TODO: think if we can remove this restriction
+        static_assert(std::is_same_v<typename SrcMapping::RecordDim, typename DstMapping::RecordDim>);
```
j-stephan: Maybe add an error message here.

include/llama/Copy.hpp
```diff
@@ -0,0 +1,381 @@
...
+    template <typename SrcMapping, typename SrcBlob, typename DstMapping, typename DstBlob>
+    void aosoaCommonBlockCopy(
+        const View<SrcMapping, SrcBlob>& srcView,
+        View<DstMapping, DstBlob>& dstView,
+        bool readOpt,
+        std::size_t threadId = 0,
+        std::size_t threadCount = 1)
+    {
+        // TODO: think if we can remove this restriction
+        static_assert(std::is_same_v<typename SrcMapping::RecordDim, typename DstMapping::RecordDim>);
```
j-stephan: And here.

include/llama/Copy.hpp
```diff
@@ -0,0 +1,381 @@
...
+        static constexpr auto LanesDst = internal::aosoaLanes<DstMapping>;
+
+        if (srcView.mapping.arrayDims() != dstView.mapping.arrayDims())
+            throw std::runtime_error{"Array dimensions sizes are different"};
+
+        static constexpr auto srcIsAoSoA = LanesSrc != std::numeric_limits<std::size_t>::max();
+        static constexpr auto dstIsAoSoA = LanesDst != std::numeric_limits<std::size_t>::max();
+
+        static_assert(srcIsAoSoA || dstIsAoSoA, "At least one of the mappings must be an AoSoA mapping");
+        static_assert(!srcIsAoSoA || decltype(srcView.storageBlobs)::rank == 1);
```
j-stephan: And here.

include/llama/Copy.hpp
```diff
@@ -0,0 +1,381 @@
...
+        static_assert(!dstIsAoSoA || decltype(dstView.storageBlobs)::rank == 1);
```
j-stephan: And here.

include/llama/Copy.hpp
```diff
@@ -0,0 +1,381 @@
...
+            }
+        }
+    }
+
+    /// @brief Generic implementation of \ref copy defaulting to \ref fieldWiseCopy. LLAMA provides several
+    /// specializations of this construct for specific mappings. Users are encouraged to also specialize this template
+    /// with better copy algorithms for further combinations of mappings, if they can and want to provide a better
+    /// implementation.
+    template <typename SrcMapping, typename DstMapping, typename SFINAE = void>
+    struct Copy
```
j-stephan, line 290: Would it make sense to hide these `Copy` structs in a separate namespace? Otherwise the user might be inclined to use them directly instead of the `copy()` function.

j-stephan (changes requested):

include/llama/Copy.hpp
bernhardmgruber: How much value would such an error message be: "The source and destination record dimensions must be the same". Because this is exactly what is already written here. I wonder if I can convey any more meaningful help.

include/llama/Copy.hpp
bernhardmgruber: done

include/llama/Copy.hpp
bernhardmgruber: done

include/llama/Copy.hpp
bernhardmgruber: Not sure. The user should call `copy()`. But when the user adds specializations, the need to specialize `Copy<...>`. So it should be inside a namespace that does not look scary. I think I prefer leaving it just in `llama`. But if you have a good suggestion, I might take it :)

include/llama/Copy.hpp
j-stephan: Well, the message you just suggested is immediately human-readable. Without the message, the user possibly needs to parse a C++ template novel first.

include/llama/Copy.hpp
j-stephan: In alpaka we put these types into the the `alpaka::traits` namespace. Maybe that is an option here?

include/llama/Copy.hpp
bernhardmgruber: Which they need to parse anyway, to know which mapping exactly, and from which copy invocation the error was raised ;) But fine, I will add it.

include/llama/Copy.hpp
bernhardmgruber: I know, but `trait` is the wrong term. `Copy` is a customization point, which is pretty long to spell.

include/llama/Copy.hpp
bernhardmgruber: Added.

include/llama/Copy.hpp
bernhardmgruber: Added.

include/llama/Copy.hpp
bernhardmgruber: Thinking again about this, I would like to keep it as is for now. Is that fine for you?

j-stephan: Sure, LLAMA is your pet ;-)

PR #290, opened 2021-06-07
Showcase and discuss CRP clang-tidy

bernhardmgruber:
This PR tests the clang-tidy file from https://github.com/ComputationalRadiationPhysics/contributing/blob/master/formatting-tools/.clang-tidy including PR https://github.com/ComputationalRadiationPhysics/contributing/pull/51. It just includes the result of running `run-clang-tidy -fix`, with no additional manual changes.

It includes the following changes to the current state of the CRP `.clang-tidy` file:
* do not prefix public members with `_m`
* ignore prefixing template parameters with 1 or 2 letters

### DIFF
--- a/examples/alpaka/nbody/nbody.cpp
+++ b/examples/alpaka/nbody/nbody.cpp
@@ -105,7 +105,7 @@ LLAMA_FN_HOST_ACC_INLINE auto store(FP& dst, Vec v)
 template<std::size_t Elems>
 struct VecType
 {
-    // TODO: we need a vector type that also works on GPUs
+    // TODO(bgruber): we need a vector type that also works on GPUs
 #ifndef ALPAKA_ACC_GPU_CUDA_ENABLED
     using type = Vc::SimdArray<FP, Elems>;
 #endif
@@ -177,7 +177,7 @@ struct UpdateKernel
         const auto ti = alpaka::getIdx<alpaka::Grid, alpaka::Threads>(acc)[0];
         const auto tbi = alpaka::getIdx<alpaka::Block, alpaka::Threads>(acc)[0];
 
-        // TODO: we could optimize here, because only velocity is ever updated
+        // TODO(bgruber): we could optimize here, because only velocity is ever updated
         auto pi = [&]
         {
             constexpr auto arrayDims = llama::ArrayDims<1>{Elems};
@@ -186,7 +186,7 @@ struct UpdateKernel
             constexpr auto blobAlloc = llama::bloballoc::Stack<llama::sizeOf<typename View::RecordDim> * Elems>{};
             return llama::allocView(mapping, blobAlloc);
         }();
-        // TODO: vector load
+        // TODO(bgruber): vector load
         LLAMA_INDEPENDENT_DATA
         for(auto e = 0u; e < Elems; e++)
             pi(e) = particles(ti * Elems + e);
@@ -204,7 +204,7 @@ struct UpdateKernel
                 pPInteraction<Elems>(pi(0u), sharedView(j));
             alpaka::syncBlockThreads(acc);
         }
-        // TODO: vector store
+        // TODO(bgruber): vector store
         LLAMA_INDEPENDENT_DATA
         for(auto e = 0u; e < Elems; e++)
             particles(ti * Elems + e) = pi(e);
--- a/examples/nbody/nbody.cpp
+++ b/examples/nbody/nbody.cpp
@@ -846,51 +846,51 @@ namespace manualAoSoA_manualAVX
 
 namespace manualAoSoA_Vc
 {
-    template<typename vec>
+    template<typename Vec>
     struct alignas(32) ParticleBlock
     {
         struct
         {
-            vec x;
-            vec y;
-            vec z;
+            Vec x;
+            Vec y;
+            Vec z;
         } pos, vel;
-        vec mass;
+        Vec mass;
     };
 
 
-    template<typename vec>
+    template<typename Vec>
     inline void pPInteraction(
-        vec piposx,
-        vec piposy,
-        vec piposz,
-        vec& pivelx,
-        vec& pively,
-        vec& pivelz,
-        vec pjposx,
-        vec pjposy,
-        vec pjposz,
-        vec pjmass)
+        Vec piposx,
+        Vec piposy,
+        Vec piposz,
+        Vec& pivelx,
+        Vec& pively,
+        Vec& pivelz,
+        Vec pjposx,
+        Vec pjposy,
+        Vec pjposz,
+        Vec pjmass)
     {
-        const vec xdistance = piposx - pjposx;
-        const vec ydistance = piposy - pjposy;
-        const vec zdistance = piposz - pjposz;
-        const vec xdistanceSqr = xdistance * xdistance;
-        const vec ydistanceSqr = ydistance * ydistance;
-        const vec zdistanceSqr = zdistance * zdistance;
-        const vec distSqr = EPS2 + xdistanceSqr + ydistanceSqr + zdistanceSqr;
-        const vec distSixth = distSqr * distSqr * distSqr;
-        const vec invDistCube = [&]
+        const Vec xdistance = piposx - pjposx;
+        const Vec ydistance = piposy - pjposy;
+        const Vec zdistance = piposz - pjposz;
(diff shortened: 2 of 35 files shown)


examples/vectoradd/vectoradd.cpp
```diff
@@ -80,72 +80,72 @@ namespace usellama
...
 {
     struct Vector
     {
-        FP x;
-        FP y;
-        FP z;
+        FP m_x;
+        FP m_y;
+        FP m_z;
     };
```
bernhardmgruber: I don't think the `m_` increases clarity here. I think it makes the usage worse.

examples/vectoradd/vectoradd.cpp
```diff
@@ -80,72 +80,72 @@ namespace usellama
...
         LLAMA_INDEPENDENT_DATA
-        for (std::size_t i = 0; i < PROBLEM_SIZE; i++)
+        for (std::size_t i = 0; i < problem_size; i++)
         {
-            c[i].x = a[i].x + b[i].x;
-            c[i].y = a[i].y - b[i].y;
-            c[i].z = a[i].z * b[i].z;
+            c[i].m_x = a[i].m_x + b[i].m_x;
+            c[i].m_y = a[i].m_y - b[i].m_y;
+            c[i].m_z = a[i].m_z * b[i].m_z;
```
bernhardmgruber: This looks gross now.

examples/alpaka/nbody/nbody.cpp
```diff
@@ -64,9 +64,9 @@ using Particle = llama::Record<
 
 enum Mapping
 {
-    AoS,
-    SoA,
-    AoSoA
+    ao_s,
+    so_a,
+    ao_so_a
```
bernhardmgruber: I guess we have to fix these by hand.

examples/alpaka/vectoradd/vectoradd.cpp
```diff
@@ -75,88 +75,88 @@ try
...
+    const DevAcc dev_acc(alpaka::getDevByIdx<PltfAcc>(0u));
+    const DevHost dev_host(alpaka::getDevByIdx<PltfHost>(0u));
+    Queue queue(dev_acc);
 
     // LLAMA
-    const auto arrayDims = llama::ArrayDims{PROBLEM_SIZE};
+    const auto array_dims = llama::ArrayDims{problem_size};
 
-    const auto mapping = [&]
+    const auto mapping = [array_dims]
```
bernhardmgruber: I don't think I like that check. Even though it might improve readability for small captures, if the capture list is larger, this becomes less readable. Plus, if I want to change the lambda content now, I have to also edit the capture list.

examples/nbody/nbody.cpp
```diff
@@ -175,64 +175,64 @@ namespace usellama
...
         auto operator*=(Vec v) -> Vec&
         {
-            x *= v.x;
-            y *= v.y;
-            z *= v.z;
+            m_x *= v.m_x;
+            m_y *= v.m_y;
+            m_z *= v.m_z;
             return *this;
         }
```
bernhardmgruber: Even though we have a class with member functions now, I do not think the `m_` is helpful here.

examples/nbody/nbody.cpp
```diff
@@ -257,86 +257,86 @@ namespace manualAoS
...
-            p.vel.y = dist(engine) / FP(10);
-            p.vel.z = dist(engine) / FP(10);
-            p.mass = dist(engine) / FP(100);
+            p.m_pos.m_x = dist(engine);
+            p.m_pos.m_y = dist(engine);
+            p.m_pos.m_z = dist(engine);
+            p.m_vel.m_x = dist(engine) / FP(10);
+            p.m_vel.m_y = dist(engine) / FP(10);
+            p.m_vel.m_z = dist(engine) / FP(10);
+            p.m_mass = dist(engine) / FP(100);
```
bernhardmgruber: Looks gross.

examples/nbody/nbody.cpp
```diff
@@ -423,45 +423,45 @@ namespace manualSoA
...
         }
-        plotFile << std::quoted(title) << "\t" << sumUpdate / STEPS << '\t' << sumMove / STEPS << '\n';
+        plot_file << std::quoted(title) << "\t" << sum_update / steps << '\t' << sum_move / steps << '\n';
 
         return 0;
     }
 } // namespace manualSoA
 
-namespace manualAoSoA
+namespace manual_ao_so_a
```
bernhardmgruber: Lol. That is really messed up :D But I guess we can fix this manually into `manual_aosoa`.

examples/nbody/nbody.cpp
```diff
@@ -1063,215 +1063,215 @@ namespace manualAoSoA_Vc
...
-    template <typename T>
-    const auto particleGatherScatterStrides<Vc::Vector<T>> = Vc::Vector<std::uint32_t>{Vc::IndexesFromZero}
+    template <typename TT>
+    const auto particle_gather_scatter_strides<Vc::Vector<TT>> = Vc::Vector<std::uint32_t>{Vc::IndexesFromZero}
         * std::uint32_t{sizeof(Particle) / sizeof(FP)};
 
-    template <typename T, std::size_t N>
-    const auto particleGatherScatterStrides<Vc::SimdArray<T, N>> = Vc::SimdArray<std::uint32_t, N>{Vc::IndexesFromZero}
+    template <typename TT, std::size_t TN>
+    const auto particle_gather_scatter_strides<Vc::SimdArray<TT, TN>> = Vc::SimdArray<std::uint32_t, TN>{Vc::IndexesFromZero}
```
bernhardmgruber: I think changing `T` and `N` to `TT` and `TN` does not help anyone. We could consider an exception to the prefixing rule, if template parameter names are single letters.

tests/core.cpp
```diff
@@ -45,7 +45,7 @@ TEST_CASE("prettyPrintType")
 #ifdef _WIN32
     replace_all(str, "__int64", "long");
 #endif
-    const auto ref = R"(llama::Record<
+    const auto *const ref = R"(llama::Record<
```
bernhardmgruber: I think this is a bit noisy now.

bernhardmgruber (commented):
Most of the stuff is really good, but there are some changes I would like to discuss.

examples/vectoradd/vectoradd.cpp
j-stephan: This seems like a good case for `// NOLINT(readability-identifier-naming)` (behind each line).

examples/nbody/nbody.cpp
j-stephan: I think LLAMA is a special case with regard to `struct` / `class` members. It might make sense to disable these checks entirely for the project.

examples/alpaka/vectoradd/vectoradd.cpp
bernhardmgruber: It turns out this is actually also caused by the identifier renaming:
```
/mnt/c/dev/llama/examples/alpaka/vectoradd/vectoradd.cpp:83:16: warning: invalid case style for local constant 'arrayDims' [readability-identifier-naming]
    const auto arrayDims = llama::ArrayDims{PROBLEM_SIZE};
               ^~~~~~~~~
               array_dims
...
/mnt/c/dev/llama/examples/alpaka/vectoradd/vectoradd.cpp:85:27: note: FIX-IT applied suggested code changes
    const auto mapping = [&]
                          ^
```
I guess in clang's internal representation the `arrayDims` variable is used by the reference capture and thus the check tries to rewrite that. I don't think there is anything we can do about this than manually reverting that one line.

tests/core.cpp
bernhardmgruber: alpaka VC today: other people think this is not noisy.

examples/alpaka/nbody/nbody.cpp
bernhardmgruber: alpaka VC: we should either add exceptions for these, since they are terms of art, or manually rename then to lower case identifiers (which should not trigger the check).

bernhardmgruber:
So far I can reasonably continue living with the proposed changeset. A last change I would consider is having global `constexpr` constants in ALLCAPS, maybe even all `constexpr` constants. I think having them `snake_case` like everything else adds a bit of confusion since they are not distinguishable from local variables and other stuff that potentially changes.

Issue #296, opened 2021-06-09
1D resizeable container


bernhardmgruber:
`std::vector<T>` is probably the most used standard data structure and a great building block for many programs. LLAMA's views, while very powerful and almost providing similar functionality as `std::vector<T>`, lack the ability to grow in size. However that is easy to implement by creating a new view when we run out of storage and copy the old content to the new view. The building blocks are there.

Such a facility is already needed by the raycasting example (not yet merged).
We also likely need a `std::deque<T>` equivalent for the integration of LLAMA into PIConGPU to represent the particle lists.

bernhardmgruber:
We now have `llama::Vector<Mapping>`, with some basic functionality.

bernhardmgruber:
I think the status quo is good enough and demonstrates that LLAMA can be used as a building block for higher-level data structures, including resizeable ones.

PR #300, opened 2021-06-10
add Vector, a 1D container

bernhardmgruber:
Addresses #296.

The implementation is not complete but contains a reasonably usable subset of what `std::vector` offers.

For the unit tests, I started with https://github.com/MFHava/PTL/blob/master/test/vector.cpp, but the current state is far different now. @MFHava would you consider the tests/vector.cpp in this PR still derived work from your original source? If yes, I would leave the license header as is and still need to add the Boost license file somewhere.

### DIFF
--- a/include/llama/ArrayDimsIndexRange.hpp
+++ b/include/llama/ArrayDimsIndexRange.hpp
@@ -222,7 +222,7 @@ namespace llama
         }
 
     private:
-        ArrayDims<Dim> size;
+        ArrayDims<Dim> size; // TODO: we only need to store Dim - 1 sizes
         ArrayDims<Dim> current;
     };
 
--- a/include/llama/Vector.hpp
+++ b/include/llama/Vector.hpp
@@ -1,0 +1,305 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+
+#pragma once
+
+#include "View.hpp"
+#include "VirtualRecord.hpp"
+
+#include <algorithm>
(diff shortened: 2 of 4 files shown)


MFHava:
> @MFHava would you consider the tests/vector.cpp in this PR still derived work from your original source?

Skimming through your unit tests, I don't see anything that would necessitate that - is there even one line of code from me remaining? If there isn't, just go ahead and claim authorship...

bernhardmgruber:
> > @MFHava would you consider the tests/vector.cpp in this PR still derived work from your original source?
> 
> Skimming through your unit tests, I don't see anything that would necessitate that - is there even one line of code from me remaining? If there isn't, just go ahead and claim authorship...

There is probably no line remaining unchanged. There are things like the default ctor check:
```c++
const ptl::vector<int> v0;
REQUIRE(v0.empty());
```
That became:
```c++
const Vector v;
CHECK(v.empty());
```
So there are a few lines left which carry on the idea. But the big portion is work of my own. Reading wikipedia on [derivative work](https://en.wikipedia.org/wiki/Derivative_work), which is by no means legal advice, but maybe a good guidance, it says:

> a derivative work is an expressive creation that includes major copyrightable elements of an original, previously created first work

I therefore would like to conclude that my unit tests are not derivative work since close to zero elements were taken over. @MFHava is that assessment correct in your opinion?

MFHava:
Yeah, just go ahead...

PR #306, opened 2021-06-24
Add MacOS CI

### LINKED ISSUE #297: Add MacOS CI job
@SimeonEhrig pointed out that LLAMA is not tested on MacOS. It should run in principle, so we should add a CI job for this.

Also update the badge on the README.md then.

bernhardmgruber:
Fixes #297.

This PR adds a CI job for MacOS. Several fixed were needed on the way. Most prominently, the `SharedPtr` blob allocator needs to use `boost::shared_ptr` now instead of `std::shared_ptr`, because libc++ < 11 (MacOS < 11) does not yet implement array support for `std::shared_ptr`.


bernhardmgruber:
It turns out MacOS 11 is not yet available for everyone on GitHub: https://github.com/actions/virtual-environments/blob/main/docs/macos-11-onboarding.md

PR #307, opened 2021-06-25
Add [push|pop]_[front|back] for Array

bernhardmgruber:
Add `[push|pop]_[front|back]` for `Array`.

### DIFF
--- a/include/llama/Array.hpp
+++ b/include/llama/Array.hpp
@@ -89,6 +89,46 @@ namespace llama
 
     template <typename First, typename... Args>
     Array(First, Args... args) -> Array<First, sizeof...(Args) + 1>;
+
+    template <typename T, std::size_t N>
+    LLAMA_FN_HOST_ACC_INLINE constexpr auto push_front(Array<T, N> a, T v) -> Array<T, N + 1>
+    {
+        Array<T, N + 1> r{};
+        r[0] = v;
+        for (auto i = 0; i < N; i++)
+            r[i + 1] = a[i];
+        return r;
+    }
+
+    template <typename T, std::size_t N>
+    LLAMA_FN_HOST_ACC_INLINE constexpr auto push_back(Array<T, N> a, T v) -> Array<T, N + 1>
+    {
+        Array<T, N + 1> r{};
+        for (auto i = 0; i < N; i++)
+            r[i] = a[i];
+        r[N] = v;
+        return r;
+    }
+
+    template <typename T, std::size_t N>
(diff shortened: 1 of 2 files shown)


include/llama/Array.hpp
```diff
@@ -89,6 +89,56 @@ namespace llama
 
     template <typename First, typename... Args>
     Array(First, Args... args) -> Array<First, sizeof...(Args) + 1>;
+
+    template <typename T, std::size_t N>
+    LLAMA_FN_HOST_ACC_INLINE constexpr auto push_front(Array<T, N> a, T v) -> Array<T, N + 1>
```
j-stephan, line 94: Is passing by value intended here?

include/llama/Array.hpp
```diff
@@ -89,6 +89,56 @@ namespace llama
...
+    LLAMA_FN_HOST_ACC_INLINE constexpr auto pop_back(Array<T, N> a) -> Array<T, N - 1>
+    {
+        Array<T, N - 1> r{};
+        for (auto i = 0; i < N - 1; i++)
+            r[i] = a[i];
+        return r;
+    }
+
+    template <typename T>
+    LLAMA_FN_HOST_ACC_INLINE constexpr auto pop_back(Array<T, 0> a) -> Array<T, 0>
```
j-stephan: What are the intended semantics here? To me this sounds like a case for `static_assert`.

j-stephan (approved):
Looking good, a few questions:

include/llama/Array.hpp
bernhardmgruber: Yes. While generally passing large arrays per value might cause expensive copies, the arrays in LLAMA are quite small usually. Most oftenly, these arrays hold the index space dimensions so usually just have 1 or maybe 2 or 3 elements. Furthermore, the call is forced inline as well, so a function call should never occur.

include/llama/Array.hpp
bernhardmgruber: I pulled this out of a different PR, where I need this precise semantic. But we could argue it should be a differently named function. I will probably change the name then.

bernhardmgruber: I found a way to not need this special case. I removed it.

include/llama/Array.hpp
j-stephan: Okay. Another question: What is the intended use case? In the test you just pass in some rvalues. Is this always the case in the real world? If not, is it intended to have two arrays after the operation?

bernhardmgruber: The way it's currently used is to add a `size_t` to an array of `size_t`s. The old and the new array continue living.

PR #308, opened 2021-06-25
Arbitarily cat and compare record coords

bernhardmgruber:
* add `operator==` and `operator!=` for `RecordCoord`
* support arbitrary many record coords in `Cat` and `cat`

### DIFF
--- a/include/llama/RecordCoord.hpp
+++ b/include/llama/RecordCoord.hpp
@@ -96,15 +96,15 @@ namespace llama
     template <typename L>
     using RecordCoordFromList = internal::mp_unwrap_sizes<L>;
 
-    /// Concatenate two \ref RecordCoord.
-    template <typename RecordCoord1, typename RecordCoord2>
-    using Cat = RecordCoordFromList<boost::mp11::mp_append<typename RecordCoord1::List, typename RecordCoord2::List>>;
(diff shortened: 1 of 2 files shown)


include/llama/RecordCoord.hpp
```diff
@@ -78,15 +96,15 @@ namespace llama
     template <typename L>
     using RecordCoordFromList = internal::mp_unwrap_sizes<L>;
 
-    /// Concatenate two \ref RecordCoord.
-    template <typename RecordCoord1, typename RecordCoord2>
-    using Cat = RecordCoordFromList<boost::mp11::mp_append<typename RecordCoord1::List, typename RecordCoord2::List>>;
+    /// Concatenate a set of \ref RecordCoord.
```
j-stephan, line 99: You can use the plural form in Doxygen like this:

```suggestion
    /// Concatenate a set of \ref RecordCoord%s.
```

j-stephan (approved):

include/llama/RecordCoord.hpp
bernhardmgruber: That is great to know, I already wondered! I guess I should review the entire doxygen for this! I will do this in a later pass: #311

PR #309, opened 2021-06-25
Improve Tuple

bernhardmgruber:
* allow construction from convertible types
* remove single element Tuple specialization
* add get() overload returning mutable reference
* implement C++ tuple protocol
* add equality comparison operators
* add a few tests

### DIFF
--- a/include/llama/Tuple.hpp
+++ b/include/llama/Tuple.hpp
@@ -9,15 +9,12 @@
 
 namespace llama
 {
-    /// Tuple class like `std::tuple` but suitable for use with offloading devices like GPUs.
     template <typename... Elements>
-    struct Tuple;
-
-    template <>
-    struct Tuple<>
+    struct Tuple
     {
     };
 
+    /// Tuple class like `std::tuple` but suitable for use with offloading devices like GPUs.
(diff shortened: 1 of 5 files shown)


include/llama/Tuple.hpp
```diff
@@ -27,58 +26,96 @@ namespace llama
...
-
-        constexpr Tuple() = default;
-
-        LLAMA_FN_HOST_ACC_INLINE
-        constexpr Tuple(FirstElement const first, Tuple<> const rest = Tuple<>()) : first(first)
-        {
-        }
-
-        FirstElement first;
+        [[no_unique_address]] RestTuple rest; ///< the remaining elements
```
j-stephan: Did LLAMA switch to C++20? The description still says its C++17.

j-stephan (approved):

include/llama/Tuple.hpp
bernhardmgruber: LLAMA compiles and works fine with C++17. However, there are several additional features available with C++20, hidden behind `#ifdef` etc. Unknown attributes in particular, like this one in C++17 mode, will just be ignored by the compiler, so we are fine.

PR #319, opened 2021-07-02
PIC example

bernhardmgruber:
Here is a Particle-In-Cell (PIC) example for LLAMA, developed by @sliwowitz and converted to LLAMA by me. The pure conversion from the non-LLAMA to the LLAMA version changed the runtime of each step from 23ms to 20ms, probably due to alignment.

TODO:

- [x] Should we keep the non-LLAMA version for comparison? No.
- [x] Embed the `input.txt` parameter file?
- [x] @sliwowitz is the implementation actually correct? Yes.
- [x] Find a better memory mapping resulting in a more impressive speedup. Maybe later.
- [x] ~~SIMD version~~
- [x] alpaka version

~~Alternatively, there is a 2D PIC implementation in the Parallel Research Kernels: https://github.com/ParRes/Kernels~~


bernhardmgruber:
This example might show the desired effect that different mappings are needed on different accelerators. In this PIC example the mapping for the particle array seems largely irrelevant. But for the 2D XYZ fields, AoS is faster an CPUs and SoA is faster on my GPU.

examples/pic/pic.cpp
```diff
@@ -0,0 +1,637 @@
...
+constexpr Real B0z = 0.1; // TODO: input.txt had 0
+constexpr Real v_drift = 0.2;
+constexpr Real v_therm = 0.005;
+constexpr Real alpha = 0.8;
+constexpr int NSTEPS = 2000;
+constexpr int outputInterval = 10;
+constexpr Real Rcx = R_ * dx + 0;
+constexpr Real Rcy = R_ * dx + 0;
+constexpr Real Rcz = 0;
+constexpr auto reportInterval = 10;
```
bernhardmgruber: @sliwowitz could you please cross check these parameters if they are correct? I also would like to have a more heavy simulation, with around 5x to 10x more work. I also think we might have too little particles. There are only 2000 particles on a 1000x1000 grid, which is the reason why the particle kernels have almost no impact on the runtime. How much can we increase the particle count? What is a realistic scenario? Thx!

bussmann:
Please also test on hemera!

bernhardmgruber:
> Please also test on hemera!

Here is a super preliminary result on hemera:

```
Tesla V100-SXM2-32GB
Particle mapping: AoS
Field mapping:    AoS
         |  steps |  clr J | integr |  dep J |  bnd J | adv B1 | bnd B1 |  adv E |  bnd E | adv B2 | bnd B2 |  total |
.........|     10 |   2250 |     21 |     17 |     27 |   2606 |     23 |   2631 |     23 |   2605 |     22 |  10229 |
Particle mapping: AoS
Field mapping:    SoA
         |  steps |  clr J | integr |  dep J |  bnd J | adv B1 | bnd B1 |  adv E |  bnd E | adv B2 | bnd B2 |  total |
.........|     10 |   2185 |     21 |     16 |     29 |   2555 |     24 |   2580 |     23 |   2516 |     23 |   9976 |

Intel(R) Xeon(R) Gold 6136 CPU @ 3.00GHz
Particle mapping: AoS
Field mapping:    AoS
         |  steps |  clr J | integr |  dep J |  bnd J | adv B1 | bnd B1 |  adv E |  bnd E | adv B2 | bnd B2 |  total |
.........|     10 |  12000 |     59 |     42 |     69 |  10561 |     60 |  13844 |     66 |  10555 |     60 |  47321 |
Particle mapping: AoS
Field mapping:    SoA
         |  steps |  clr J | integr |  dep J |  bnd J | adv B1 | bnd B1 |  adv E |  bnd E | adv B2 | bnd B2 |  total |
.........|     10 |  12131 |     59 |     43 |    104 |  10543 |     97 |  14241 |     91 |  10550 |     95 |  47958 |
```

It shows the average of 10 steps for the individual kernels of the PIC simulation when the fields are stored as AoS vs SoA. SoA is 1.025x faster on GPU, AoS is 1.013x faster on CPU. That is a very minor difference, but shows that different layouts do perform better on different architectures. However, the simulation does not run with realistic parameters. I am waiting on @sliwowitz to update those.

bussmann:
I think this is a minute difference not supporting your claim. I wonder in what way a choice in parameters would change this.

bernhardmgruber:
For smaller problem sizes I can start making a case here:
image
image
The differences flatten out with larger problem sizes.
I hope the difference will be even more dramatic with the incoming fixes from @sliwowitz.
I would like to test on hemera as well, but I cannot get a slot for the last two days.

bussmann:
Underutilizing the GPU is a weak case. I think you should think of a better, more illustrative example.

examples/pic/pic.cpp
bernhardmgruber: Discussed offline.

bernhardmgruber:
Benchmarks on NVIDIA P100, NVIDIA V100, NVIDIA RTX 2060, then Intel Xeon, AMD EPYC, AMD Ryzen
4002x4002 grid, 12.8M particles
image
Unfortunately, the runtimes of the different memory layouts are less diverse between the top and the bottom benchmarks than what I was hoping for.

bussmann:
The bigger GPUs are surely not fully utilized with these numbers.

bussmann:
The thing you are doing is always just AoS vs. SoA, but complex algorithms require more special data layouts & careful tuning. You either find an illustrative example that is of real world value where AoS vs. SoA makes sense or you dig deep into the need of specialized memory layouts. Given the time, I'd siggest the former.

bernhardmgruber:
Here are more memory layout variations:
image
image
Even though the morton array layout for the field is the best on the GPU, taking the P: SoA F: AoS combination is very close to optimal on the GPU and the best on the CPU.

Here are the individual kernels that make up a simulation step:
image

Still seems like I need to find a different example for the paper.

examples/alpaka/pic/pic.cpp
```diff
@@ -0,0 +1,960 @@
...
+template<typename VirtualParticle, typename FieldView>
+LLAMA_FN_HOST_ACC_INLINE void deposit_current(const VirtualParticle& part, FieldView& J)
+{
+    /* Calculation of previous position */
+    const Real igamma = 1 / sqrt(1 + squaredNorm(part(U{})));
+    const auto r_old = part(R{}) - igamma * dt * part(U{}); // Position should be a 2D vector
+
+    /* Particle position to cell coordinates */
+    const auto i = static_cast<size_t>(part(R{}, X{}) * inv_dx);
+    const auto i_shift = static_cast<size_t>(part(R{}, X{}) * inv_dx + 0.5);
```
bernhardmgruber: @sliwowitz please figure out whether the `+0.5` and cast is correct. clang-tidy flags this and suggests `lround` if rouding `part(R{}, X{}) * inv_dx` to the nearest integer is intended.

bussmann:
You are beating a dead horse with this approach. A PIC data layout is very special, too special for what you are trying to achieve in such short time. Let's talk!

examples/alpaka/pic/pic.cpp
bernhardmgruber: I replaced this by `lround` now.

PR #323, opened 2021-07-05
Align and minimize One

### LINKED ISSUE #267: One with minimum padding
`llama::One<Record>` is a facility for creating a scalar record value backed by stack memory. With the padding minimization developed in #233, we could add an additional version of `One` that reorders it's members to minimize padding.

Such a feature is not available in C++ and was traditionally done manually, see e.g.: http://www.catb.org/esr/structure-packing/
There is an effort to get such a feature in C++: https://wg21.link/p1112. LLAMA would be a library solution to this.
Rust has this built-in.

Such a one could actually just be used by itself, e.g. `std::vector<llama::MinOne<T>>`.

bernhardmgruber:
Fixes #267.

### DIFF
--- a/include/llama/Array.hpp
+++ b/include/llama/Array.hpp
@@ -38,12 +38,12 @@ namespace llama
         LLAMA_FN_HOST_ACC_INLINE constexpr T* end()
         {
             return &element[N];
-        };
+        }
 
         LLAMA_FN_HOST_ACC_INLINE constexpr const T* end() const
         {
             return &element[N];
-        };
+        }
 
         template <typename IndexType>
         LLAMA_FN_HOST_ACC_INLINE constexpr auto operator[](IndexType&& idx) -> T&
@@ -91,6 +91,53 @@ namespace llama
         }
     };
 
+    template <typename T>
+    struct Array<T, 0>
+    {
+        using value_type = T;
+        static constexpr std::size_t rank = 0;
+
+        LLAMA_FN_HOST_ACC_INLINE constexpr auto size() const
+        {
+            return 0;
+        }
+
+        LLAMA_FN_HOST_ACC_INLINE constexpr T* begin()
+        {
+            return nullptr;
+        }
+
+        LLAMA_FN_HOST_ACC_INLINE constexpr const T* begin() const
+        {
+            return nullptr;
+        }
+
+        LLAMA_FN_HOST_ACC_INLINE constexpr T* end()
+        {
+            return nullptr;
+        }
+
+        LLAMA_FN_HOST_ACC_INLINE constexpr const T* end() const
+        {
+            return nullptr;
+        }
+
+        LLAMA_FN_HOST_ACC_INLINE constexpr friend auto operator==(const Array& a, const Array& b) -> bool
+        {
+            return true;
+        }
+
+        LLAMA_FN_HOST_ACC_INLINE constexpr friend auto operator!=(const Array& a, const Array& b) -> bool
+        {
+            return false;
+        }
+
+        LLAMA_FN_HOST_ACC_INLINE constexpr friend auto operator+(const Array& a, const Array& b) -> Array
+        {
+            return {};
+        }
+    };
+
     template <typename First, typename... Args>
     Array(First, Args... args) -> Array<First, sizeof...(Args) + 1>;
 
@@ -99,8 +146,9 @@ namespace llama
     {
         Array<T, N + 1> r{};
         r[0] = v;
-        for (auto i = 0; i < N; i++)
-            r[i + 1] = a[i];
+        if constexpr (N > 0)
+            for (auto i = 0; i < N; i++)
+                r[i + 1] = a[i];
         return r;
     }
 
@@ -108,8 +156,9 @@ namespace llama
     LLAMA_FN_HOST_ACC_INLINE constexpr auto push_back(Array<T, N> a, T v) -> Array<T, N + 1>
     {
         Array<T, N + 1> r{};
-        for (auto i = 0; i < N; i++)
-            r[i] = a[i];
+        if constexpr (N > 0)
+            for (auto i = 0; i < N; i++)
+                r[i] = a[i];
         r[N] = v;
         return r;
     }
@@ -119,8 +168,9 @@ namespace llama
     {
         static_assert(N > 0);
         Array<T, N - 1> r{};
-        for (auto i = 0; i < N - 1; i++)
-            r[i] = a[i];
+        if constexpr (N > 1)
+            for (auto i = 0; i < N - 1; i++)
+                r[i] = a[i];
         return r;
     }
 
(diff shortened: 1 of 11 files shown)


include/llama/Array.hpp
```diff
@@ -86,6 +86,48 @@ namespace llama
         }
     };
 
+    template <typename T>
+    struct Array<T, 0>
```
j-stephan, line 95: Wouldn't an array with dimensionality 0 contain exactly one element? If we imagine it geometrically it would be the equivalent to a point.

include/llama/Array.hpp
```diff
@@ -94,17 +136,19 @@ namespace llama
     {
         Array<T, N + 1> r{};
         r[0] = v;
-        for (auto i = 0; i < N; i++)
-            r[i + 1] = a[i];
+        if constexpr (N > 0)
```
j-stephan, line 149: We probably want some curly braces behind the `if`. I think some compilers even warn about this. Also in the other instances if you change this.

include/llama/mapping/One.hpp
```diff
@@ -4,12 +4,21 @@
...
+#include "Common.hpp"
 
 namespace llama::mapping
 {
     /// Maps all ArrayDims coordinates into the same location and layouts struct members consecutively. This mapping is
     /// used for temporary, single element views.
-    template <typename T_ArrayDims, typename T_RecordDim>
+    /// \tparam AlignAndPad If true, padding bytes are inserted to guarantee that struct members are properly aligned.
+    /// If false, struct members are tighly packed.
+    /// \tparam FlattenRecordDim Defines how the record dimension's fields should be flattended. See \ref
```
j-stephan: Typo: 
```suggestion
    /// \tparam FlattenRecordDim Defines how the record dimension's fields should be flattened. See \ref
```

include/llama/mapping/One.hpp
```diff
@@ -4,12 +4,21 @@
...
 #include "../Core.hpp"
+#include "Common.hpp"
 
 namespace llama::mapping
 {
     /// Maps all ArrayDims coordinates into the same location and layouts struct members consecutively. This mapping is
     /// used for temporary, single element views.
-    template <typename T_ArrayDims, typename T_RecordDim>
+    /// \tparam AlignAndPad If true, padding bytes are inserted to guarantee that struct members are properly aligned.
+    /// If false, struct members are tighly packed.
```
j-stephan: Typo:
```suggestion
    /// If false, struct members are tightly packed.
```

j-stephan (changes requested):

include/llama/Array.hpp
bernhardmgruber: That's an array of *size* 0 :)
`llama::Array` is equivalent to slimmed down `std::array`.

include/llama/Array.hpp
bernhardmgruber: That's a stylistic question, which I strongly oppose. I know the argument that adding braces could avoid accidentially misindenting a statement, believing it belongs to the `if` statement. However, clang actually detects such misalignment. Furthermore, our code base is clang-formatted so we would immediately notice. So the only thing braces add here is an extra LOC, spreading the code a little vertically. And I prefer to the code to be compact. I also know clang-tidy can add these braces automatically, so at some point we will have the vote and then we can see what happens :)

include/llama/mapping/One.hpp
bernhardmgruber: Thx!

include/llama/mapping/One.hpp
bernhardmgruber: thx!

include/llama/mapping/AoS.hpp
```diff
@@ -12,7 +12,7 @@ namespace llama::mapping
     /// If false, struct members are tightly packed.
     /// \tparam T_LinearizeArrayDimsFunctor Defines how the array dimensions should be mapped into linear numbers and
     /// how big the linear domain gets.
-    /// \tparam FlattenRecordDim Defines how the record dimension's fields should be flattended. See \ref
+    /// \tparam FlattenRecordDim Defines how the record dimension's fields should be flatteneded. See \ref
```
j-stephan: Typo: 
```suggestion
    /// \tparam FlattenRecordDim Defines how the record dimension's fields should be flattened. See \ref
```

include/llama/mapping/One.hpp
```diff
@@ -4,12 +4,21 @@
...
+#include "Common.hpp"
 
 namespace llama::mapping
 {
     /// Maps all ArrayDims coordinates into the same location and layouts struct members consecutively. This mapping is
     /// used for temporary, single element views.
-    template <typename T_ArrayDims, typename T_RecordDim>
+    /// \tparam AlignAndPad If true, padding bytes are inserted to guarantee that struct members are properly aligned.
+    /// If false, struct members are tightly packed.
+    /// \tparam FlattenRecordDim Defines how the record dimension's fields should be flatteneded. See \ref
```
j-stephan: Typo:
```suggestion
    /// \tparam FlattenRecordDim Defines how the record dimension's fields should be flattened. See \ref
```

include/llama/mapping/One.hpp
```diff
@@ -35,14 +44,39 @@ namespace llama::mapping
...
+    /// One mapping preserving the alignment of the field types by inserting padding.
+    /// \see One
+    template <typename ArrayDims, typename RecordDim>
+    using AlignedOne = One<ArrayDims, RecordDim, true, FlattenRecordDimInOrder>;
+
+    /// One mapping preserving the alignment of the field types by inserting padding and permuting the field order to
+    /// minimize this padding.
+    /// \see One
+    template <typename ArrayDims, typename RecordDim>
+    using MinAlignedOne = One<ArrayDims, RecordDim, true, FlattenRecordDimMinimizePadding>;
```
j-stephan, line 76: Is `One<ArrayDims, RecordDim, false, FlattenRecordDimMinimizePadding>` missing on purpose?

include/llama/mapping/One.hpp
```diff
@@ -35,14 +44,39 @@ namespace llama::mapping
...
+
+    /// One mapping packing the field types tightly, violating the types alignment requirements.
```
j-stephan:
```suggestion
    /// One mapping packing the field types tightly, violating the types' alignment requirements.
```

j-stephan (changes requested):

include/llama/mapping/One.hpp
bernhardmgruber: That is a valid mapping, but I do not see the use case. `One<..., true, ...>` adds padding to respect alignment, so adding field permutations make sense, to play with this padding.
`One<..., false, ...>` does not add padding, so permuting the fields has a far smaller effect. So I want to advertise it less.

include/llama/mapping/AoS.hpp
bernhardmgruber: Thx!

include/llama/mapping/One.hpp
bernhardmgruber: Thx!

include/llama/mapping/One.hpp
bernhardmgruber: Thx!

j-stephan (approved):

PR #345, opened 2021-08-30
Allow [[no_unique_address]] on nvcc 11.4

bernhardmgruber:


bernhardmgruber:
Looks like nvcc 11.4 with gcc/clang still ICEs :/

PR #349, opened 2021-08-30
[on hold] Add thrust/PSTL example with particles and tracks

bernhardmgruber:
TODO

This PR includes hacks to LLAMA for which we still need to find better solutions.
We also use a custom fork of thrust to fix supporting proxy references.


bernhardmgruber:
We could consider rebasing this example just on the parallel STL and use it with nvc++.

PR #351, opened 2021-08-31
Dop support for Intel's classic compiler (icp)

### LINKED ISSUE #350: Remove support for icp
Intel is starting to deprecate their classic compiler (icp) and advertises customers to move to their new LLVM based compiler (icpx): https://software.intel.com/content/www/us/en/develop/blogs/adoption-of-llvm-complete-icx.html.
LLAMA has support for both in its CI. However, maintaining icp is sometimes additional work because it offers a unique experience of quirks. We should remove support for icp.

bernhardmgruber:
Addresses #350.


j-stephan (approved):
3wfmft

PR #355, opened 2021-08-31
Sufficiently align blobs

### LINKED ISSUE #339: llama::One is insufficiently aligned
`llama::One` represents and maintains storage for a single record value. From the perspective of the compiler however, it is just a byte array plus the mapping and an empty array dimension coord. The latter two occupy no space and are optimized away by `[[no_unique_address]]`. The compiler thus computes an alignment of 1 for a `llama::One`. However, on access, bytes inside a `llama::One` are reinterpreted as types with higher than one alignment requirements. `llama::One` itself thus needs to be aligned properly depending on the contained record values.

bernhardmgruber:
This PR fixes #339, by extending the blob allocator interface to also receive an alignment. This alignment is computed in `allocView`.
The new APIs `alignOf` and `flatAlignOf` are added as well.

### DIFF
--- a/docs/pages/blobs.rst
+++ b/docs/pages/blobs.rst
@@ -7,8 +7,14 @@ Blobs
 
 When a :ref:`view <label-view>` is created, it needs to be given an array of blobs.
 A blob is an object representing a contiguous region of memory where each byte is accessible using the subscript operator.
-The number of blobs and the size of each blob is a property determined by the mapping used by the view.
+The number of blobs and the alignment/size of each blob is a property determined by the mapping used by the view.
 All this is handled by :cpp:`llama::allocView()`, but I needs to be given a blob allocator to handle the actual allocation of each blob.
+
+.. code-block:: C++
+
+    auto blobAllocator = ...;
+    auto view = llama::allocView(mapping, blobAllocator);
+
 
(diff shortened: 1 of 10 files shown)


docs/pages/blobs.rst
```diff
@@ -20,43 +26,38 @@ Contrary, if a :cpp:`std::shared_ptr<std::byte[]>` is used, the storage is share
...
-A blob allocator is a callable which returns an appropriately sized blob given a desired allocation size in bytes.
-There is a number of a buildin blob allocators:
+A blob allocator is a callable which returns an appropriately sized blob given a desired compile-time alignment and runtime allocation size in bytes.
+Choosing the right compile-time alignment has implications on the read/write speed on some CPU architectures and may even lead to CPU exceptions if data is not properly aligned.
+A blob allocator is called like this:
+
+.. code-block:: C++
+
+    auto blobAllocator = ...;
+    auto blob = blobAllocator(std::integral_constant<std::size_t, Alignment>{}, size);
```
bernhardmgruber, line 36: I am not entirely sure I like the design. Passing the `integral_constant` is quite verbose. But I also do not want to have a template parameter for it because then we cannot use lambda's as blob allocators in C++17.

PR #360, opened 2021-09-13
SIMD

### LINKED ISSUE #169: Implement a SIMD for_each
LLAMA is a library for efficient memory layouts and memory access patterns. For CPUs, this is only half the story. The second half is filling vector registers to also optimize the compute throughput. While this should not  be directly handled by LLAMA, LLAMA certainly can provide useful data layouts and should allow for some algorithmic primitives on top.

After thinking some time on this topic and rereading some stuff from Matthias Kretz on Vc and std::simd, I think these primitives are going to be the SIMD variants of the STL algorithms, as proposed here: http://wg21.link/P0350.

Modified example from P0350:
```c++
template <class T>
T something(T); // T can by float or std::simd<float>

auto f(const std::vector<float>& data) {
  std::vector<float> output;
  stdx::transform(std::execution::simd,
    data.begin(), data.end(), output.begin(),
    [](auto x) { // type of x can by float or std::simd<float>
      return something(x + 1);
    });
  return output;
}
```

I think just having a `for_each` in this fassion could already be quite useful. Even more useful, if we would allow the lambda argument to be a mutable reference allowing write-back to memory. Depending on the data layout and target architecture, we could then run the lambda with SIMD vectors or scalars.

### LINKED ISSUE #295: Some<Record> a construct for some elements and SIMD
LLAMA has the `One<Record>` construct to hold and manipulate a single record. To facilitate SIMD programming models, we would need a construct to hold several records, depending on the target hardware architecture and the given record dimension. We should explore adding a `Some<Record>` construct representing a bunch of records.

Whether `Some<Record>` should expose a size is an open question: `Some<Record, N>`.

`Some<Record>` for CPU targets should create a structure of SIMD vectors and thus should integrate with SIMD libraries like Vc or `std::simd`. But it should neither require a SIMD library (GPU targets) or become a SIMD library (mission creep). The interaction is mostly on how to store from any mapping into SIMD vectors. Whether that is efficient or not depends on the mapping (e.g. vector fetches vs. gather/scatter vs. individual loads).

A design aspect is how to differentiate a scalar access from a vector access. This information also needs to be carried through the VirtualRecord's created on the way until the access is terminal.

I believe this construct could solve the inefficiency of the AoSoA mapping, which is slow when individual elements are accessed. But when `Some<Records>`s are read from an AoSoA with matching lane length, access could be a lot better. Furthermore, this construct is a good building block on how algorithms could be designed to be SIMD friendly and a stepping stone towards a vectorizable kernel language.

bernhardmgruber:
This PR adds prototype APIs for SIMD programming with LLAMA.

It includes the following facilities:
* simdLanes
* SimdizeN
* Simdize
* simdLanesFor
* SimdN<T, N, MakeSizedSimd>
* Simd<T, MakeSimd> (#295)
* loadSimd
* storeSimd
* simdForEachN
* simdForEach (#169)


bernhardmgruber:
n-body update/move using `llama::Some`:
image
The SIMD variants are in most cases optimal.

j-stephan:
What does `MB` stand for in the tests?

bernhardmgruber:
> What does `MB` stand for in the tests?

MultiBlob.

bussmann:
How do you define optimal? Is there any baseline to which you compare to?

bernhardmgruber:
> How do you define optimal? Is there any baseline to which you compare to?

I compare to the non-SIMD versions and inspect the generated disassembly.

bernhardmgruber:
So, and this is where it gets super annoying. I cannot reproduce the failure of nvcc 11.2 and nvcc 11.5 locally, neither on my workstation nor the workstation of a coworker ...

bernhardmgruber:
I could narrow down the LOC which causes nvcc to die, but it is just an ordinary access on a `RecordRef`. I decided to disable the example now for nvcc <= 11.5.

bernhardmgruber:
image

For LLAMA SoA SIMD, we are looking good. LLAMA AoS SIMD and LLAMA AoSoA SIMD are still too slow. I suspect the issue is in the generated vector load/stores, which have no special handling in LLAMA yet.

bernhardmgruber:
image

The alpaka n-body for CPU looks good as well, it vectorizes accordingly and the assembly for the SoA produces the right instructions:
image

PR #393, opened 2021-10-05
Test macos-11 and macos-10.15 in CI

bernhardmgruber:
This PR splits the macOS CI into 2, running on macOS 10.15 and macOS 11.

GitHub seems to have changed `macos-latest` to point to `macos-11`, where the compiler crashes at the moment. Since we don't have a corresponding machine to reproduce, I will make the macOS 11 CI optional for now.


bernhardmgruber:
Tracking macOS 11 crash here: #394.

PR #403, opened 2021-10-19
add LLAMA_UNROLL

bernhardmgruber:
This PR adds the macro `LLAMA_UNROLL(n)`, requesting the compiler to unroll the following loop `n` times.


bernhardmgruber:
It looks like we cannot get this working entirely with g++, because it does not accept values from template arguments. See also: https://stackoverflow.com/q/63404539/2406044.

PR #420, opened 2021-11-29
Move bitpack mappings into LLAMA

### LINKED ISSUE #184: Integer packing and arbitrary precision integers
LLAMA could allow to pack integers into fewer bits than their usual size. Especially memory footprint sensitive data layouts frequently use such types to save memory.
Another approach is the support of arbitrary precision integer types, which are also common in FPGA code, e.g.  `ap_int<N>`: https://www.xilinx.com/html_docs/xilinx2020_2/vitis_doc/use_arbitrary_precision_data_type.html

E.g.: 3 12-bit integers forming an RGB value:
```c++
using RecordDim = llama::Record<
    llama::Field<R, llama::Int<12>>,
    llama::Field<G, llama::Int<12>>,
    llama::Field<B, llama::Int<12>>,
>;
```

An open design point is how a reference to such an object is formed, since a mapping may not place these objects at a byte-boundary. Thus, locations of such elements might not be addressable. A solution could be a proxy reference like in e.g. `std::bitset<N>`.

bernhardmgruber:


bernhardmgruber:
Superseded by #427

PR #426, opened 2021-12-08
Switch to 2 threads for building

bernhardmgruber:
GitHub runners seem to have 2 cores per job: https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners.


bernhardmgruber:
It turns out that the builds are now slower than before. Abandoning.

PR #442, opened 2021-12-14
Add Null mapping

bernhardmgruber:
The Null mappings maps all elements to nothing. Writing data through a reference obtained from the Null mapping discards the value. Reading through such a reference returns a default constructed object.


bussmann:
Next up: Identity mapping.

bernhardmgruber:
> Next up: Identity mapping.

I am not sure what you have in mind here. How can I have an identity mapping, when domain (data space) and codomain (blob nr/offset space) are different sets?

PR #448, opened 2021-12-18
Simplify CI

bernhardmgruber:


bernhardmgruber:
Ok, we need to fix this with Boost in one go, because we cannot make the CI succeed without the Boost upgrade.
Superceded by: #447.

PR #476, opened 2022-02-28
Disable MSVC workaround for OpenMP pragmas in lambdas

bernhardmgruber:
This is supposedly fixed in Visual Studio 2022 17.2: https://developercommunity.visualstudio.com/t/OpenMP-in-lambda-expression-compile-erro/1501041

- [x] Verify that this is actually fixed once Visual Studio 2022 17.2 is available in the GitHub CI.


bernhardmgruber:
Just checked, the CI is still using MSVC 19.31 and not 19.32, where this should be fixed.

bernhardmgruber:
CI just passes the tests with MSVC 19.32.

PR #488, opened 2022-05-09
Support extents with arbitrary value types

bernhardmgruber:
Currently, LLAMA uses `size_t` in all index calculations as is the default in the STL. We knew already from previous experience in picongpu that this increases register usage for CUDA since `size_t` is 64bits in CUDA and CUDA registers are 32bits in size, so each index variable uses 2 registers. Registers are a scarce resource so it would be beneficial to have control over the type used in index calculations. This PR allows this.

We saw similar development in `std::mdspan` (https://wg21.link/P0009) with the proposal “Make mdspan size_type controllable” (https://wg21.link/P2553). The gist was adding an additional template parameter to `std::extents` which gives the type used in all index calculations:

```c++
template<class SizeT, size_t... Extents>
  class extents;
```

For LLAMA, the main change is that `llama::ArrayExtents` gains an additional template parameter defining the type to use for storing extents and performing index calculations in:
```c++
using AE = llama::ArrayExtents<IndexType, 5, llama::dyn, 2>;
```

This change is propagated downstream into `llama::ArrayExtentsDynamic<IndexType, 3>`, which is a breaking change. `llama::NrAndOffset` is now templated on the index type as well. Furthermore `ArrayExtentsStatic` is renamed to `ArrayExtentsNCube`, since this is a more accurate description of what it does.


bernhardmgruber:
Just tested this one in AdePT and it works.

PR #498, opened 2022-05-13
Use bash for downloading boost on Windows

bernhardmgruber:


bernhardmgruber:
Too hard to setup, aborting.

PR #516, opened 2022-06-01
Add vcpkg manifest and simplify CI

bernhardmgruber:


bernhardmgruber:
Abandoned in favor of #700. The vcpkg version requirements and baselines are too complicated for now.

PR #521, opened 2022-06-07
Make structName() and recordCoordTags() constexpr and use in Trace with CUDA

bernhardmgruber:


bernhardmgruber:
Unfortunately, I cannot reproduce the Windows build failure locally :/

PR #533, opened 2022-06-15
CI fixes

bernhardmgruber:
* Explicitly set CUDA host compiler to the compiler used in the CI for building the non-CUDA examples
* Remove the g++-12/nvcc 11.7 runs, as g++-12 is not yet supported by nvcc.


bernhardmgruber:
This should uncover that the gcc12 runs actually fail with CUDA.

PR #547, opened 2022-06-30
Support nvc++

### LINKED ISSUE #203: Add nvc++ to the CI
@ax3l mentioned a couple of relevant HPC compilers [here](https://github.com/alpaka-group/alpaka/issues/1296#issuecomment-823861871). LLAMA already tests with all publicly available ones, except for the new nvc++ from NVIDIA.

nvc++ recently started to support CUDA as well, so it can be used as a host and device compiler (much like clang). We should add it to the CI.

bernhardmgruber:
So I picked up nvc++ again after returning from [PASC22](https://pasc22.pasc-conference.org/) since a few people reported on quite positive experiences with it. This PR adds nvc++ to the CI and adds the necessary changes to LLAMA.


bernhardmgruber:
The necessary changes to LLAMA are surprisingly limited so far. A big blocker is Vc though, which causes:
`NVC++-S-0000-Internal compiler error. ?? unexpected constant 0 (/usr/local/include/Vc/sse/../common/storage.h: 256)`

PR #562, opened 2022-08-30
Retry downloading xsimd

bernhardmgruber:
Installing the head of xsimd via vcpkg fails regularly, so let's run it in a loop until it succeeds.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -13,7 +13,7 @@ env:
   THREADS: 4
   CONFIG: RelWithDebInfo
   ALPAKA_BRANCH: 0.9.0
-  VCPKG_INSTALL: "vcpkg install fmt tinyobjloader boost-mp11 boost-atomic boost-smart-ptr boost-functional boost-container; vcpkg install --head xsimd"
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -13,7 +13,7 @@ env:
   THREADS: 4
   CONFIG: RelWithDebInfo
   ALPAKA_BRANCH: 0.9.0
-  VCPKG_INSTALL: "vcpkg install fmt tinyobjloader boost-mp11 boost-atomic boost-smart-ptr boost-functional boost-container; vcpkg install --head xsimd"
+  VCPKG_INSTALL: "vcpkg install fmt tinyobjloader boost-mp11 boost-atomic boost-smart-ptr boost-functional boost-container; until vcpkg install --head xsimd; do echo 'Retrying'; done"
```
PrometheusPi, line 16: @bernhardmgruber might this cause an infinite loop or does there exist an external timeout?

bernhardmgruber: AFAIK the github action will be terminated after some time if this never succeeds.

PR #565, opened 2022-09-01
Enforce identifier naming

bernhardmgruber:
The names used in LLAMA have grown too inconsistent over time. Mostly a mix of `camelBack`, `CamelCase` and `snake_case` is used. This PR enables several more identifier naming checks in clang-tidy and enforces them in the CI.

PR #566, opened 2022-09-02
Fix inconsistencies in clang-tidy CI

bernhardmgruber:


bernhardmgruber:
It seems we also need to merge the fixes from #565 at the same time to pass the CI.

bernhardmgruber:
This PR was included in #565

PR #568, opened 2022-09-05
Add dump tests for computed mappings

bernhardmgruber:
Merge first:

- [x] #569


bernhardmgruber:
It seems the codecov uploader has an issue. I am trying to fix the issue here: #569

PR #592, opened 2022-10-31
Fix vcpkg commit to work around broken Catch2 on MacOS

bernhardmgruber:


bernhardmgruber:
The MacOS runner is going to be fixed upstream by GitHub: https://github.com/actions/runner-images/pull/6490
The case is documented here #594 until the issue is fixed.

PR #610, opened 2022-11-10
Fix formatting only on push

bernhardmgruber:


bernhardmgruber:
I am actually happy with the status quo, so I will abandon this.

PR #648, opened 2022-12-02
Align SoA subarrays when using single blob

bernhardmgruber:
This changeset will align SoA sub arrays by default when using a single blob. This can cause a potential performance loss, but improves correctness.

### DIFF
--- a/include/llama/mapping/SoA.hpp
+++ b/include/llama/mapping/SoA.hpp
@@ -21,10 +21,12 @@ namespace llama::mapping
         Align
     };
 
-    /// Struct of array mapping. Used to create a \ref View via \ref allocView.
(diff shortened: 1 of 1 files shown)


psychocoderHPC:
Why could this result in a performance loss?

bernhardmgruber:
Because I need to do additional arithmetic in the mapping function which depends on the array size to compute the necessary padding. If the array size is runtime, this padding can also only be computed at runtime. Furthermore, you need to repeat that computation for each struct member that comes before the one you are accessing. It is essentially this loop: https://github.com/alpaka-group/llama/blob/develop/include/llama/mapping/SoA.hpp#L119-L127. With #653, I can precompute a table and avoid the loop iff the array extents are known at compile time.

PR #650, opened 2022-12-07
Add Babelstream

bernhardmgruber:
This PR adds Babelstream from alpaka and wraps the data into LLAMA views. This way, we can show whether LLAMA would introduce any overhead.


bernhardmgruber:
image

Using LLAMA, we have the same throughput as without, so there is no immediate overhead.

PR #651, opened 2022-12-08
add HIP support

psychocoderHPC:
- add function attribute to define macros for HIP
- add unroll support for HIP
- fix missing function defines for function friend declarations

### DIFF
--- a/include/llama/macros.hpp
+++ b/include/llama/macros.hpp
@@ -34,7 +34,7 @@
 #endif
 
 #ifndef LLAMA_FORCE_INLINE
-#    if defined(__NVCC__)
+#    if defined(__NVCC__) || defined(__HIP__)
 #        define LLAMA_FORCE_INLINE __forceinline__
 #    elif defined(__GNUC__) || defined(__clang__)
 #        define LLAMA_FORCE_INLINE inline __attribute__((always_inline))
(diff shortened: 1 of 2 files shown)


include/llama/macros.hpp
```diff
@@ -52,7 +52,8 @@
 #endif
 
 #ifndef LLAMA_UNROLL
-#    if defined(__NVCC__) || defined(__NVCOMPILER) || defined(__clang__) || defined(__INTEL_LLVM_COMPILER)
+#    if defined(__HIP__) || defined(__NVCC__) || defined(__NVCOMPILER) || defined(__clang__)                          \
+        || defined(__INTEL_LLVM_COMPILER)
```
bernhardmgruber, line 56: Since `__HIP__` is also handled via `__clang__` here, either remove the `__HIP__` here or add it to the definitions of other macros in this file (e.g. `LLAMA_FORCE_INLINE`).

bernhardmgruber (changes requested):

include/llama/macros.hpp
psychocoderHPC: I added it to `LLAMA_FORCE_INLINE` too

bernhardmgruber: Great, thx!

bernhardmgruber (approved):

PR #653, opened 2022-12-09
Improve SoA with static array extents

bernhardmgruber:
This PR improves the SoA by precomputing a compile-time only subarray offset table when the array extents are fully static.

This PR also contains an extension of the vectoradd benchmarks:
    * Mix data types in record dimension, so padding and alignment can have an effect
    * Make problem size an odd number
    * Plot results
    * Test all LLAMA versions in a single run
    * Test a few variations of the SoA layout
    * Simplify alpaka use


bernhardmgruber:
image

Runtime difference is marginal:

| layout  | runtime  |
|---|---|
| "SoA SB aligned"	| 0.113894 |
| "SoA SB aligned CT size" | 0.113254 |

compile time:
image

runtime:

image

bernhardmgruber:
image

layout | runtime
-- | --
"SoA SB aligned" | 0.160085
"SoA SB aligned CT size" | 0.160387

Here, the CT variant is actually a bit slower. But the assembly is better (less instructions).

PR #660, opened 2022-12-19
Add ROOT lhcb analysis example

bernhardmgruber:
This example is based on the following TTree/RNTuple benchmark: https://github.com/jblomer/iotools/blob/master/lhcb.cxx

The example reads experimental physics data through RNTuple into a LLAMA view, runs an exemplary analysis, fills a historgram and displays the result.

TODO:

- [x] Test/Run the executable. It currently does not work because of a version missmatch in the input data: https://github.com/jblomer/iotools/issues/9

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -148,6 +148,12 @@ if (LLAMA_BUILD_EXAMPLES)
 		message(WARNING "Could not find alpaka. Alpaka examples are disabled.")
 	endif()
 
+	# ROOT examples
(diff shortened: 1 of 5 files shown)


bernhardmgruber:
There is a workaround for generating up-to-date RNTuple's, as described in https://github.com/jblomer/iotools/issues/9#issuecomment-1374935754. Let's test that and then merge the example.

bernhardmgruber:
The result is now the same as the corresponding analysis from @jblomer's iotools:

image

PR #672, opened 2023-01-11
Study layout effects on LHCB B2HHH analysis

bernhardmgruber:
This changeset extends the ROOT LHCB B2HHH analysis example:

* Parallelize histogram generation
* Test multiple memory layouts
* Save histograms to disk
* Extend benchmark result by mean, stddev, error

### DIFF
--- a/examples/root/lhcb_analysis/CMakeLists.txt
+++ b/examples/root/lhcb_analysis/CMakeLists.txt
@@ -2,9 +2,11 @@ cmake_minimum_required (VERSION 3.18)
 project(llama-root-lhcb_analysis)
 
 find_package(ROOT REQUIRED)
+find_package(OpenMP REQUIRED)
 if (NOT TARGET llama::llama)
 	find_package(llama REQUIRED)
 endif()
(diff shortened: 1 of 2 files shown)


bernhardmgruber:
Conversion vs. analysis time:
image

Dataset sizes:
image

bernhardmgruber:
Here is an animation of the impact of used mantissa bits on the observable:
bitpacked_mantissa_animation
And exponent bits:
bitpacked_exponent_animation

bernhardmgruber:
AoS heatmap of first 10000 events:
image

bernhardmgruber:
AoS layout visualization:
image

image
Where the Custom layout bitpacks H1isMuon/H2isMuon/H3isMuon as AoS, 1 bit each, into a blob of their own, then groups H1ProbK and H2ProbK as AoS in a separate blob, then the rest as AoS in another blob. The Custom layout has been built up gradually using insights obtained via instrumentation and benchmarking.

bernhardmgruber:
Data set sizes:
image

PR #696, opened 2023-01-30
Replace zenodo.json by CITATION.cff

### LINKED ISSUE #337: Add CITATION.cff
It seems there is a new emerging way to specify meta data to cite software: https://citation-file-format.github.io/
This could generalize/replace the zenodo.json. And it has GitHub integration, giving users an easy way to get a BibTex entry.

bernhardmgruber:
Fixes: #337


.zenodo.json
```diff
@@ -1,54 +0,0 @@
-{
```
psychocoderHPC, line 1: wait, this is not possible zenodo.json is used to publish alpaka and requred for the webhook!
The author list in this file is requred ro auto generate the author list for the doi entry.

psychocoderHPC (changes requested):
I do not think that we can remove the zenodoo file

.zenodo.json
bernhardmgruber: No, when you have a `CITATION.cff`, zenodo will take this one and ignore the `.zenodo.json`. See here: https://github.com/citation-file-format/citation-file-format/issues/374#issuecomment-1061022832

psychocoderHPC: I was not aware that this works, that's nice.
We need to test the release workflow in another project first to avoid that we create a wrong release that is not removable in zenodoo.

psychocoderHPC (approved):

bernhardmgruber:
I ran the `CITATION.cff` through the [validator](https://citation-file-format.github.io/cff-initializer-javascript/#/) again and fixed a few things.

PR #700, opened 2023-02-03
Add vcpkg manifest without dependency versions

bernhardmgruber:
All LLAMA dependencies will now be installed automatically via vcpkg by cmake.


bernhardmgruber:
This should be unblocked now by my upstream fix in vcpkg: [#34072](https://github.com/microsoft/vcpkg/pull/34072)

PR #751, opened 2023-07-21
Improve concept definitions

bernhardmgruber:
This PR formats and improves some of the concept definitions in LLAMA.


bernhardmgruber:
The concept formatting looks way nicer with clang-format-16. I think we should update before merging this PR.

Edit: done in #752

PR #754, opened 2023-07-27
Disable a SIMD testing failing in GH

bernhardmgruber:
I fail to reproduce the test locally. All tests pass with ASAN locally.


bernhardmgruber:
Just as I thought, disabling the offending SIMD test on g++-12 and g++-13 just makes the tests crash in another part of the unit tests.

PR #784, opened 2023-10-01
Rename linearizers towards mdspan jargon

bernhardmgruber:
* Rename all linearizers from `LinearizeArrayDims...` to `LinearizeArrayIndex...`
* Rename `...Cpp` to `...Right`, but leave alias for compatibility with legacy. This increases compatibility with mdspan.
* Rename `...Fortran` to `...Left`, but leave alias for compatibility with legacy. This increases compatibility with mdspan.

### DIFF
--- a/docs/pages/api.rst
+++ b/docs/pages/api.rst
@@ -191,11 +191,11 @@ RecordDim field permuters
 Common utilities
 ^^^^^^^^^^^^^^^^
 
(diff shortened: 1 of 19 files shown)


bernhardmgruber:
The Intel apt repository key expired, so I am ignoring the corresponding CI job.

PR #786, opened 2023-10-02
Align AoSoA sub arrays

bernhardmgruber:
This was not supported before, but is necessary for correctness. AoSoA now gets the same alignment parameter as AoS.

- [x] Measure if there is a performance difference


bernhardmgruber:
Tested with n-body against develop, no difference.

PR #796, opened 2023-11-16
Add a compile time benchmark

bernhardmgruber:
This adds a small benchmark with configureable record dimension size, so we can benchmark the impact of record dimension size on the compilation time.


bernhardmgruber:
image

